PR #16, opened 2023-12-13
boardswarm-cli: Rework ui to use ratatui crate

obbardc:
The tui crate is depreciated and ratatui is the continuation project of tui. Prepare for future tui additions by converting to use the ratatui crate instead.


sjoerdsimons:
@obbardc ooi why is this in draft?

obbardc:
@sjoerdsimons I have now undrafted it. I suppose I was going to add more commits on top to use additional ratatui features - but that can always come in more PRs in future - i don't have immediate plans to do that :-)

PR #27, opened 2024-02-07
Improve help and diagnostic messages

andrzejtp:


boardswarm/src/config.rs
```diff
@@ -93,6 +94,7 @@ pub struct ModeStep {
 
 impl Config {
     pub fn from_file<P: AsRef<Path>>(path: P) -> Result<Config> {
+        info!("Trying configuration file {}", path.as_ref().display());
```
obbardc, line 97: You may wish to drop this chunk in favour of https://github.com/boardswarm/boardswarm/pull/30

obbardc (changes requested):

boardswarm-cli/src/main.rs
```diff
@@ -389,7 +389,7 @@ struct ConfigureArg {
 
 #[derive(Debug, Subcommand)]
 enum Command {
-    /// Configure boardswarm instances
+    /// Configure boardswarm instances access
```
sjoerdsimons, line 411: This change is handled in #23 so best to drop it from this PR to avoid conflicts later on

boardswarm/src/config.rs
sjoerdsimons: They do both serve a slightly different purpose ofcourse; This shows in the logs with config file is used not just if it failed ;) Though i'd probably be more optimistic and word it as `Loading` rather then trying but that's not a merge blocker for me

boardswarm-cli/src/main.rs
```diff
@@ -389,61 +408,71 @@ struct ConfigureArg {
...
+        /// Type of items to monitor
+        type_: ItemTypes,
     },
     /// Show item properties
     Properties {
-        #[arg(value_parser = parse_item)]
-        type_: ItemType,
+        #[arg(value_enum)]
+        /// Type of items to show properties of
+        type_: ItemTypes,
```
sjoerdsimons, line 468: This is net progress so again not a merge blocker for me; But one thing that has annoyed me at times is that for the properties command the type should really be singular not plural (e.g. `properties devices 1`) makes no sense.  Though for properties it's possible to refactor into the item types specific commands at least and have (`device 1 properties`); monitor doesn't really fit into that mold though so we should keep that.

boardswarm-cli/src/main.rs
```diff
@@ -35,6 +35,44 @@ mod ui;
...
+        match val {
+            ItemTypes(t) => t,
+        }
+    }
+}
+
+impl std::fmt::Debug for ItemTypes {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let t: ItemType = (*self).into();
+        write!(f, "{:?}", t)
```
sjoerdsimons, line 52:
```suggestion
        write!(f, "{:?}", self.0)
```

boardswarm-cli/src/main.rs
```diff
@@ -35,6 +35,44 @@ mod ui;
...
 mod utils;
 
+#[derive(Clone, Copy)]
+struct ItemTypes(pub ItemType);
+
+impl From<ItemTypes> for ItemType {
+    fn from(val: ItemTypes) -> Self {
+        match val {
+            ItemTypes(t) => t,
+        }
```
sjoerdsimons, line 45:
```suggestion
        val.0
```

boardswarm-cli/src/main.rs
```diff
@@ -35,6 +35,44 @@ mod ui;
...
+        ]
+    }
+
+    fn to_possible_value(&self) -> Option<PossibleValue> {
+        Some(match self {
+            ItemTypes(ItemType::Actuator) => PossibleValue::new("actuators"),
+            ItemTypes(ItemType::Console) => PossibleValue::new("consoles"),
+            ItemTypes(ItemType::Device) => PossibleValue::new("devices"),
+            ItemTypes(ItemType::Volume) => PossibleValue::new("volumes"),
+        })
```
sjoerdsimons, line 72:
```suggestion
        Some(match self.0 {
            ItemType::Actuator => PossibleValue::new("actuators"),
            ItemType::Console => PossibleValue::new("consoles"),
            ItemType::Device => PossibleValue::new("devices"),
            IItemType::Volume => PossibleValue::new("volumes"),
        })
```

boardswarm-cli/src/main.rs
```diff
@@ -389,61 +408,71 @@ struct ConfigureArg {
...
 #[derive(Debug, Subcommand)]
 enum Command {
-    /// Configure boardswarm instances
+    /// Configure boardswarm instances access
     Configure(ConfigureArg),
     /// Retrieve the login information from the remote boardswarm server
     LoginInfo,
     /// Actuator specific commands
     Actuator {
+        /// Which actuator to use
```
sjoerdsimons, line 417: The convention that i seems to have been using has been `The X to` e.g:
```
$ boardswarm-cli  device write --help
Write data to a device volume

Usage: boardswarm-cli device <DEVICE> write [OPTIONS] <VOLUME> <TARGET> <FILE>

Arguments:
  <VOLUME>  The volume to write to
  <TARGET>  The volume target to write to
  <FILE>    Path to the file to write
```
Would be good to follow a consistent convention in all help messages

boardswarm-cli/src/main.rs
sjoerdsimons, line 52: Or potentially even better just directly call out to the wrapped types implementation so you avoid a potential useless str conversion 
```suggestion
        self.0.fmt(f)
```

sjoerdsimons (changes requested):
A bunch of suggestions but primarily it would be good to drop the first commit and leave that to #23 :)

sjoerdsimons (approved):

PR #31, opened 2024-02-09
gpio: Print debug information about unmatched devices & add tracing instrumentation to start_provider

obbardc:
gpio: Print debug information about unmatched devices
    
It can be helpful to understand why a gpio device wasn't registered. Rework
the registration code to check if it matches against the required device
earlier, then print the device's name and properties if the matching
fails e.g:
    
    DEBUG boardswarm::gpio: Ignoring gpio device /sys/devices/platform/AMDI0030:00/gpiochip0 - Properties { properties: {"udev.DEVNAME": "/dev/gpiochip0", ... }

### DIFF
--- a/boardswarm/src/dfu.rs
+++ b/boardswarm/src/dfu.rs
@@ -7,6 +7,7 @@ use tokio::sync::{
     mpsc::{self, Receiver},
     oneshot,
 };
+use tracing::instrument;
 use tracing::{info, warn};
 
 use crate::{
@@ -14,6 +15,7 @@ use crate::{
 };
 pub const PROVIDER: &str = "dfu";
 
+#[instrument(skip(server))]
 pub async fn start_provider(name: String, server: Server) {
     let provider_properties = &[
         (registry::PROVIDER_NAME, name.as_str()),
--- a/boardswarm/src/gpio.rs
+++ b/boardswarm/src/gpio.rs
@@ -3,6 +3,7 @@ use std::{collections::HashMap, path::PathBuf};
 use futures::StreamExt;
 use serde::Deserialize;
 use tokio_gpiod::{Chip, Lines};
+use tracing::instrument;
 use tracing::{debug, warn};
 
 use crate::{
@@ -37,6 +38,7 @@ impl GpioParameters {
     }
 }
 
+#[instrument(skip(parameters, server))]
 pub async fn start_provider(name: String, parameters: serde_yaml::Value, server: Server) {
     let provider_properties = &[
         (registry::PROVIDER_NAME, name.as_str()),
--- a/boardswarm/src/gpio.rs
+++ b/boardswarm/src/gpio.rs
@@ -3,7 +3,7 @@ use std::{collections::HashMap, path::PathBuf};
 use futures::StreamExt;
 use serde::Deserialize;
 use tokio_gpiod::{Chip, Lines};
-use tracing::{info, warn};
+use tracing::{debug, warn};
 
 use crate::{
     registry::{self, Properties},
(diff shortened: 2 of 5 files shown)


sjoerdsimons:
I don't think we shold stick the full debug properties at info! level; that seems ok for debug! or  likely trace! -- Though the log should be a admin tool, it should really not be a configuration helper :p

 for the use-case when configuring the recommended way is to use the `boardswarm-cli` `monitor` and `properties` to ask the server about the properties; potentially the `monitor` functionality in the cli could grow an option to also retrieve and print the properties on Add

obbardc:
ah I only did this to match the gpiochip debugging

I will try the monitor commands listed, they look helpful!

sjoerdsimons:
> ah I only did this to match the gpiochip debugging
> 
> I will try the monitor commands listed, they look helpful!

Ah right; yes looks like i left a bit too much early debugging in there :) the prints in the providers are potentilally not as useful anyway as the registration will also create a message; 

ooi are you planning to tune this further or would you prefer to close it for now ?

sjoerdsimons (changes requested):
as discussed in conversation

obbardc:
ill modify this pr to remove extra debugging in gpio  :-)

obbardc:
OK i misread you; so it looks like `boardswarm monitor` would need a lot of rework to print out the properties - how would you feel about keeping the log messages but lowering them to `debug` in the meantime ?

sjoerdsimons:
> OK i misread you; so it looks like `boardswarm monitor` would need a lot of rework to print out the properties of not yet registered devices - how would you feel about keeping the log messages but lowering them to `debug` in the meantime ?

for the lower levels; that's what i said in my initial comment right ;) I'd be fine with that; Fwiw I'd like `info` to be a good default log-level for boardswarm, so putting out a reasonable amount of info but not going into full detailed debug spew. Where that balance lies is always tricky but especially with udev based properties there tend to be a *lot* of them so i feel that's in the latter category :)

wrt. boardswarm monitor; the changes to allow it to optionally print out the properties should be quite simple :). At this point in time i mostly do this manually, watch the new device pop up in the log or monitor then run `properties` to dump out the properties if needed

boardswarm/src/rockusb.rs
```diff
@@ -48,6 +48,13 @@ pub async fn start_provider(name: String, server: Server) {
                     Ok(rockusb) => {
                         let mut properties = d.properties(name);
                         properties.extend(provider_properties);
+
+                        debug!(
+                            "potential rockusb device detected: {} - {:?}",
```
sjoerdsimons, line 53: nothing potential about this; the literal next action registers it ;)

boardswarm/src/serial/mod.rs
```diff
@@ -41,6 +41,13 @@ pub async fn start_provider(name: String, server: Server) {
                         let console = SerialPort::new(path);
                         let mut properties = d.properties(name);
                         properties.extend(provider_properties);
+
+                        info!(
+                            "potential serial device detected: {} - {:?}",
```
sjoerdsimons, line 46: ditto this is not a potential one

boardswarm/src/serial/mod.rs
```diff
@@ -41,6 +41,13 @@ pub async fn start_provider(name: String, server: Server) {
                         let console = SerialPort::new(path);
                         let mut properties = d.properties(name);
                         properties.extend(provider_properties);
+
+                        info!(
```
sjoerdsimons, line 45: debug! or trace! please

boardswarm/src/gpio.rs
```diff
@@ -56,11 +56,9 @@ pub async fn start_provider(name: String, parameters: serde_yaml::Value, server:
                         let name = name.to_string_lossy().into_owned();
                         let mut properties = d.properties(name);
 
-                        info!(
-                            "GPIO?: {} - {} - {} - {:?}",
+                        debug!(
+                            "potential gpio device detected: {} - {:?}",
                             d.syspath().display(),
-                            registration.is_some(),
-                            properties.matches(&parameters.match_),
```
sjoerdsimons, line 63: You made me remember how this works again; the registration is not needed anymore indeed (this block wouldn't execute if there is a registration anyway) however the `properties.matches` is useful; Though it probably makes sense to debug  in the lower if/else block so it doesn't get evaluated twice; Essentially it would be good to get a debug message indicate *this* provider is ignoring a certain gpio device as it doesn't match 

Fwiw you'd also probably want to throw in an `#[instrument]` at the top to at least create a span with the name of the provider so it's clear *which* gpio provider is ignoring a given device  as i don't think that's currently included

boardswarm/src/dfu.rs
```diff
@@ -42,6 +42,13 @@ pub async fn start_provider(name: String, server: Server) {
                 let dfu = Dfu::new(busnum, devnum).await;
                 let mut properties = d.properties(name);
                 properties.extend(provider_properties);
+
+                debug!(
+                    "potential dfu device detected: {} - {:?}",
```
sjoerdsimons, line 47: Nothing potential here if the next function call registers the device :)

sjoerdsimons (changes requested):

boardswarm/src/gpio.rs
obbardc: fwiw I am looking at how the gpio stuff works currently; this is where this PR came from and the reason why I put "potential" everywhere else and just confused myself :-)

So let me fix that as per what you have said here.

sjoerdsimons (approved):
fwiw the instrument should be more consistent throughout most likely to make sure for all messages we get the relevant provided logged :) but this is a great start

PR #33, opened 2024-02-10
Warn when match configuration is empty

obbardc:
If the `match` configuration is empty, we are (by-design) quite greedy and match any subdevice. Warn the user about that situation.

### DIFF
--- a/boardswarm/src/gpio.rs
+++ b/boardswarm/src/gpio.rs
@@ -38,13 +38,17 @@ impl GpioParameters {
     }
 }
 
-#[instrument(skip(parameters, server))]
+#[instrument(fields(name), skip_all, level = "error")]
 pub async fn start_provider(name: String, parameters: serde_yaml::Value, server: Server) {
     let provider_properties = &[
         (registry::PROVIDER_NAME, name.as_str()),
         (registry::PROVIDER, PROVIDER),
     ];
     let parameters: GpioParameters = serde_yaml::from_value(parameters).unwrap();
+    if parameters.match_.is_empty() {
+        warn!("matches is empty - will match any gpio device");
+    }
+
     let mut registration = None;
     let mut devices = crate::udev::DeviceStream::new("gpio").unwrap();
     while let Some(d) = devices.next().await {
--- a/boardswarm/src/main.rs
+++ b/boardswarm/src/main.rs
@@ -20,7 +20,7 @@ use thiserror::Error;
 use tokio::sync::{broadcast, mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::Streaming;
-use tracing::{info, warn};
+use tracing::{info, instrument, warn};
 
 mod boardswarm_provider;
 mod config;
@@ -239,19 +239,31 @@ trait DeviceConfigItem {
 }
 
 impl DeviceConfigItem for config::Console {
+    #[instrument(fields(name = self.name), skip_all, level="error")]
     fn matches(&self, properties: &Properties) -> bool {
+        if self.match_.is_empty() {
+            warn!("Console matches is empty - will match any console");
+        }
         properties.matches(&self.match_)
     }
 }
 
 impl DeviceConfigItem for config::Volume {
+    #[instrument(fields(name = self.name), skip_all, level="error")]
     fn matches(&self, properties: &Properties) -> bool {
+        if self.match_.is_empty() {
+            warn!("Volume matches is empty - will match any volume");
+        }
         properties.matches(&self.match_)
     }
 }
 
 impl DeviceConfigItem for config::ModeStep {
+    #[instrument(skip_all, level = "error")]
     fn matches(&self, properties: &Properties) -> bool {
+        if self.match_.is_empty() {
+            warn!("ModeStep matches is empty - will match any device");
+        }
         properties.matches(&self.match_)
     }
 }
--- a/boardswarm/src/registry.rs
+++ b/boardswarm/src/registry.rs
@@ -36,6 +36,10 @@ impl Properties {
         self.properties.get(prop).map(String::as_ref)
     }
 
+    /// Tests if matches is a subset of the properties
+    ///
+    /// If properties is from a remote instance (`boardswarm.instance` is set) that has to be
+    /// explicitly matched otherwise it's a pure subset match (e.g. an empty set matches)
     pub fn matches<K, V, I>(&self, matches: I) -> bool
     where
         K: AsRef<str>,
--- a/boardswarm/src/registry.rs
+++ b/boardswarm/src/registry.rs
@@ -264,6 +264,9 @@ mod test {
         t.insert(NAME.to_string(), "test".to_string());
         assert!(props.matches(&t));
 
+        let empty: HashMap<String, String> = HashMap::new();
+        assert!(props.matches(empty));
+
         assert!(props.matches([(NAME, "test")]));
         assert!(props.matches([("udev.BADGER", "5")]));
 


sjoerdsimons:
properties matching an empty set is expected; What are you actually trying to fix here?

obbardc:
> properties matching an empty set is expected; What are you actually trying to fix here?

so this PR specifically fixes matching in two places (there are probably also cases where `DeviceConfigItem` is used):

1) matching gpio provider:
```
providers:
  - name: gpio0
    provider: gpio
    parameters:
      match:
        # empty
      lines:
        - name: "gpio0-0"
          line_number: 0
```

2) matching a device's console :
```
devices:
  - name: test
    consoles:
      - name: main
        parameters:
          rate: 1500000
        match:
          # empty
```

If `match` is empty for either of these cases (hence `matches` is empty), `Properties::matches` will choose the last entry in the device list instead of matching. Oops!

Happy to put this check elsewhere (/rework the commit message to show the above); but I don't agree that `Properties::matches` should match an empty set from the existing uses.

sjoerdsimons:
> > properties matching an empty set is expected; What are you actually trying to fix here?
> 
> so this PR specifically fixes matching in two places (there are probably also cases where `DeviceConfigItem` is used):
> 
> 1. matching gpio provider:
> 
> ```
> providers:
>   - name: gpio0
>     provider: gpio
>     parameters:
>       match:
>         # empty
>       lines:
>         - name: "gpio0-0"
>           line_number: 0
> ```
> 
> 2. matching a device's console :
> 
> ```
> devices:
>   - name: test
>     consoles:
>       - name: main
>         parameters:
>           rate: 1500000
>         match:
>           # empty
> ```
> 
> If `match` is empty for either of these cases (hence `matches` is empty), `Properties::matches` will choose the last entry in the device list instead of matching. Oops!

Well no it *does* matching; it just matches *everything*; which means what get associated ends up being a "random" item of the expected type.

The same problem occurs if your match isn't specific enough e.g. say you match against `boardswarm.provider: serial`  then you'll get a "random" console provided by the serial provider.  While this is a silly example, but the same can happen if you match on e.g. `udev.X` properties that don't uniquely match a *specific* device. Ofcourse boardswarm cannot reasonably know what is and isn't a good match in the generic case.

To be a devils advocate, if you know your setup will only ever have *one* gpiochip or *one* console an empty match is actually ok as that uniquely identifies your unique device ! :stuck_out_tongue: 

Fundamentally you can't really fix what you're trying to fix in the generic case. But for these specific examples the providers/device handler could check for empty matches as appropriate and either `warn!` or error when a match is empty to indicate to the user this may cause unexpected behaviour (and the same should probably be emphasized in the documentation).

If you want to go one step further it probably makes sense to `warn!` when there are multiple matches in case that's not expected but that would need more refacting as currently e.g. Registry::find will just find the first matching case

> Happy to put this check elsewhere (/rework the commit message to show the above); but I don't agree that `Properties::matches` should match an empty set from the existing uses.

Properties::matches is a subset match; the empty set is a subset of everything, that's simply how set theory works.  
The code at that level really shouldn't second guess that this is probably not what you wanted, you may well have a use-case for a "match anything".

obbardc:
ok will add a check further up the chain and leave Properties::match unchanged.

boardswarm/src/registry.rs
```diff
@@ -36,6 +36,9 @@ impl Properties {
         self.properties.get(prop).map(String::as_ref)
     }
 
+    /// Tests if the elements match all of the elements of an iterator.
```
sjoerdsimons, line 39: The `warn` phrasing in your commit message is a bit strange for a pretty standard and common semantic

boardswarm/src/registry.rs
```diff
@@ -36,6 +36,9 @@ impl Properties {
...
+    ///
+    /// An empty iterator returns `true`.
```
sjoerdsimons, line 41:
```suggestion
    /// Tests if matches is a subset of properties
    ///
    /// If properties is from a remote instance (`boardswarm.instance` is set)  that has to be 
    /// explicitly matched otherwise it's a pure subset match (e.g. an empty set matches)
```

We're not passed an iterator we're passed something that can be turned into an iterator ; also i don't think i've seen

boardswarm/src/gpio.rs
```diff
@@ -45,6 +45,10 @@ pub async fn start_provider(name: String, parameters: serde_yaml::Value, server:
         (registry::PROVIDER, PROVIDER),
     ];
     let parameters: GpioParameters = serde_yaml::from_value(parameters).unwrap();
+    if parameters.match_.is_empty() {
+        warn!("matches is empty - could match random device");
```
sjoerdsimons, line 49: your commit message doesn't match your code. Also i suspect if you set your log level to warn it won't log the start_provider span (but that can be a seperate fix)

sjoerdsimons, line 49:
```suggestion
        warn!("matches is empty - will match any gpio device");
```

boardswarm/src/main.rs
```diff
@@ -239,19 +239,31 @@ trait DeviceConfigItem {
 }
 
 impl DeviceConfigItem for config::Console {
+    #[instrument(fields(name = self.name), skip(self, properties))]
```
sjoerdsimons, line 242:
```suggestion
    #[instrument(fields(name = self.name), skip_all, level="error")]
```
set the span to the highest verbosity so it's always included on output (by default it's info level). Ofcourse alternatively given this is such a small function may make sense to just put the name in the message directly.

I do wonder if this wouldn't be better in `Device::from_config` as a check on device creation (so it dumps all the warnings just once) as opposed to checks in `matches` which would warns every time a match is checked.

boardswarm/src/main.rs
```diff
@@ -239,19 +239,31 @@ trait DeviceConfigItem {
...
     fn matches(&self, properties: &Properties) -> bool {
+        if self.match_.is_empty() {
+            warn!("Console matches is empty - could match random device");
```
sjoerdsimons, line 245:
```suggestion
            warn!("Console matches is empty - will match any console");
```

boardswarm/src/main.rs
```diff
@@ -239,19 +239,31 @@ trait DeviceConfigItem {
...
+
         properties.matches(&self.match_)
     }
 }
 
 impl DeviceConfigItem for config::Volume {
+    #[instrument(fields(name = self.name), skip(self, properties))]
     fn matches(&self, properties: &Properties) -> bool {
+        if self.match_.is_empty() {
+            warn!("Volume matches is empty - could match random device");
```
sjoerdsimons, line 256:
```suggestion
            warn!("Volume matches is empty - will match any volume");
```

sjoerdsimons (changes requested):

boardswarm/src/registry.rs
obbardc: Changed from `warn` to `remind`.

boardswarm/src/registry.rs
obbardc: ok

boardswarm/src/gpio.rs
obbardc: Fixed that.

boardswarm/src/gpio.rs
```diff
@@ -38,13 +38,17 @@ impl GpioParameters {
     }
 }
 
-#[instrument(skip(parameters, server))]
+#[instrument(fields(name), skip_all, level = "error")]
```
sjoerdsimons, line 41: why  put this span at error level?

sjoerdsimons: ah i suggested that; that makes sense ;) shows how long ago it :p

boardswarm/src/main.rs
sjoerdsimons: Well you'll be seeing the behaviour of it potentially spamming the log :)  An empty match is really surprising though such i don't really mind either way

sjoerdsimons (approved):

PR #41, opened 2024-02-26
boardswarm: Add example systemd service & udev rules

obbardc:
Add an example systemd service file to the share directory.

### DIFF
--- a/boardswarm/README.md
+++ b/boardswarm/README.md
@@ -111,6 +111,10 @@ For each item created by a provider the `boardswarm.provider.name` property
 is set to the configured name and the `boardswarm.provider` property is set to
 the provider used.
 
+As a starting point, the [example udev rules](share/99-boardswarm.rules) can be
+used to grant device permissions to boardswarm. It is recommended that these
+rules be used alongside the [example systemd service](share/boardswarm.service).
+
 ### Serial provider
 
 The serial provider creates consoles from local serial ports. No provider specific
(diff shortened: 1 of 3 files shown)


sjoerdsimons:
can you add a reference to the file in the documentation as well

mwelchuk:
> For what it's worth, this systemd unit and udev rule works for me.

Spoke to soon. For reasons I needed to reboot my machine. After reboot, even with the udev rules in the other patch added, the TTYs aren't set to the correct group. I think the issue is that the udev rules are running before boardswarm is started. As a result, the (dyanmic) user doesn't actually exist and thus the required configuration isn't applied.

I think this really needs to have a system user and group created as part of install and these set in the systemd unit.

mwelchuk:
Could you also add the udev rules that would be required for DFU on TI devices?:
```
# Allow boardswarm to access TI devices when in DFU mode
ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6162", GROUP="boardswarm"
ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6165", GROUP="boardswarm"
```

obbardc:
> Could you also add the udev rules that would be required for DFU on TI devices?:
> 
> ```
> # Allow boardswarm to access TI devices when in DFU mode
> ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6162", GROUP="boardswarm"
> ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6165", GROUP="boardswarm"
> ```

Done. I will squash commits on a future next update.

PR #42, opened 2024-02-26
Add Dockerfile & build containers in CI

obbardc:
I wasn't sure if it would be best to build two images, one for the CLI and one for the server, or just one image with both binaries so I just built _something_ quickly ready for feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,3 +6,34 @@ Distributed service to interact with development boards.
 * [boardswarm-protocol](boardswarm-protocol/README.md) - Boardswarm protocol definitions
 * [boardswarm-cli](boardswarm-cli/README.md) - Boardswarm command line client
 * [boardswarm-client](boardswarm-client/README.md) - Boardswarm client library
+
+
+## Usage (Docker container)
+
+A Docker container containing all of the boardswarm binaries is available from the
+[GitHub Container Registry](https://github.com/boardswarm/boardswarm/pkgs/container/boardswarm)
+but can alternatively be built from the root of a git checkout using:
+
+```
+docker build --file boardswarm/Dockerfile --tag boardswarm .
+docker build --file boardswarm-cli/Dockerfile --tag boardswarm-cli .
+```
+
+
+Run `boardswarm-cli` inside the container:
+
+```
+docker run -it --rm \
+  ghcr.io/boardswarm/boardswarm-cli:main \
+  --help
+```
+
+
+Run `boardswarm` inside the container:
+
+```
+docker run -it --rm \
+  -p 6683:6683 \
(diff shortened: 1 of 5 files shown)


sjoerdsimons:
probably best to have different containers for cli and server; that would avoid having to jump  through hoops to run the server from it's container.

obbardc:
> probably best to have different containers for cli and server; that would avoid having to jump through hoops to run the server from it's container.

Done. Added build argument which allows choice of binary to build.

README.md
```diff
@@ -6,3 +6,36 @@ Distributed service to interact with development boards.
...
+  --help
+```
+
+
+Run `boardswarm` inside the container:
+
+```
+docker run -it --rm \
+  --port 16421:16421 \
+  --mount "type=bind,source=$(pwd)/boardswarm/share/server.conf,destination=/etc/boardswarm/server.conf" \
```
sjoerdsimons, line 37: why the very verbose mount line ooi

README.md
```diff
@@ -6,3 +6,36 @@ Distributed service to interact with development boards.
...
+  ghcr.io/boardswarm/boardswarm-cli:main \
+  --help
+```
+
+
+Run `boardswarm` inside the container:
+
+```
+docker run -it --rm \
+  --port 16421:16421 \
```
sjoerdsimons, line 36: the default port is 6683; the 16421 is the efault port of the pdudaemon

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+RUN apt update && \
+    apt install --yes \
+      libudev1 \
+      libssl3 && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY --from=builder /usr/src/boardswarm/target/release/$BINARY /usr/local/bin/$BINARY
+
+ENTRYPOINT ["/usr/local/bin/$BINARY"]
```
sjoerdsimons, line 44: for boardswarm you probably want the default entrypoint to have a default configuration path already for ease of use

README.md
```diff
@@ -6,3 +6,36 @@ Distributed service to interact with development boards.
...
+  --mount "type=bind,source=$(pwd)/boardswarm/share/server.conf,destination=/etc/boardswarm/server.conf" \
+  --entrypoint boardswarm \
```
sjoerdsimons, line 38: why is the entrypoint being overriden rather then using the default form the container

README.md
obbardc: no reason; changed to --volume now

README.md
obbardc: oops, solved

Dockerfile
obbardc: done had to separate the dockerfiles.

PR #64, opened 2024-03-13
Allow literals for actuators consoles volumes

andrzejtp:
Make e.g. this:

$ boardswarm-cli console ttyUSB2 properties

possible, also for actuators and volumes.

I am not impressed by how much parse_console(), parse_actuator() and parse_volume() are similar. Maybe room for improvement.

I've been also thinking for a while about "-c/--console" option to "ui", "device connect" and "device tail", which seems a different kind of animal. This option is specifically expected to be a string rather than a number. I am not sure if the "-c" and the additions from this PR should be integrated or not, and if they should, then how.

### DIFF
--- a/boardswarm-cli/src/main.rs
+++ b/boardswarm-cli/src/main.rs
@@ -37,7 +37,7 @@ mod ui;
 mod ui_term;
 mod utils;
 
-#[derive(Clone, Copy)]
+#[derive(Clone, Copy, Debug)]
 struct ItemTypes(pub ItemType);
 
 impl From<ItemTypes> for ItemType {
@@ -46,9 +46,24 @@ impl From<ItemTypes> for ItemType {
     }
 }
 
-impl std::fmt::Debug for ItemTypes {
+impl From<ItemType> for ItemTypes {
+    fn from(val: ItemType) -> Self {
+        Self(val)
+    }
+}
+
+impl std::fmt::Display for ItemTypes {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        self.0.fmt(f)
+        if f.alternate() {
+            std::fmt::Debug::fmt(&self.0, f)
+        } else {
+            match self.0 {
+                ItemType::Device => f.write_str("device"),
+                ItemType::Console => f.write_str("console"),
+                ItemType::Actuator => f.write_str("actuator"),
+                ItemType::Volume => f.write_str("volume"),
+            }
+        }
     }
 }
 
@@ -248,21 +263,16 @@ enum ItemArg {
     Name(String),
 }
 
-async fn item_lookup(
+async fn item_lookup<I: Into<ItemTypes>>(
     arg: ItemArg,
-    item_type: ItemType,
+    item_type: I,
     mut client: Boardswarm,
 ) -> Result<u64, anyhow::Error> {
-    let item_kind = match item_type {
-        ItemType::Device => ("Device", "device"),
-        ItemType::Console => ("Console", "console"),
-        ItemType::Actuator => ("Actuator", "actuator"),
-        ItemType::Volume => ("Volume", "volume"),
-    };
+    let item_type: ItemTypes = item_type.into();
     match arg {
         ItemArg::Id(id) => Ok(id),
         ItemArg::Name(name) => {
-            let mut items = client.list(item_type).await?;
+            let mut items = client.list(item_type.into()).await?;
 
             let (name, instance) = name
                 .rsplit_once('@')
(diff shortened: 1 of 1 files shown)


boardswarm-cli/src/main.rs
```diff
@@ -386,6 +386,54 @@ struct BmapWriteArgs {
...
 }
 
+#[derive(Clone, Debug)]
+enum VolumeArg {
+    Id(u64),
+    Name(String),
+}
+
+impl VolumeArg {
+    async fn volume(&self, mut client: Boardswarm) -> Result<Option<u64>, anyhow::Error> {
```
sjoerdsimons, line 396: This never seems to return `Ok(None)` so can just be `Result<u64, ...>` i guess.. Ok i guess this should be `lookup_volume` or `find_volume` or somesuch just to make it more clear what it does

sjoerdsimons, line 396: Looking at this and all the other lookup functions; you'd probably want to just have one function that takes an ItemType to figure out what to search for

sjoerdsimons:
> I've been also thinking for a while about "-c/--console" option to "ui", "device connect" and "device tail", which seems a different kind of animal. This option is specifically expected to be a string rather than a number. I am not sure if the "-c" and the additions from this PR should be integrated or not, and if they should, then how.

It makes no sense to integrate them; The ones in this PR are about naming Items ; while the console options refers to a *named* console on a device

sjoerdsimons (changes requested):
see comments :)

boardswarm-cli/src/main.rs
```diff
@@ -386,6 +386,54 @@ struct BmapWriteArgs {
...
+        match self {
+            VolumeArg::Id(id) => Ok(Some(*id)),
+            VolumeArg::Name(name) => {
+                let mut items = client.list(ItemType::Volume).await?;
+                items.retain(|i| i.name.starts_with(name));
```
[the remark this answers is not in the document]
andrzejtp: I thought the name contained the "@something" part, in which case instances would be taken care of: If the prefix is unique enough then it is a match. E.g. if there are

ttyUSB0
ttyUSB0@another-instance

then "ttyUSB0" is ambiguous if "ttyUSB0" and "ttyUSB0@another-instance" are the names. OTOH if there are e.g.

ttyUSB1
ttyUSB0@another-instance

then "ttyUSB0" _is_ unique enough.

The code, after matching on a prefix, counts how many matches there are. If there's just 1, then it's a perfect match, if there's 0 then nothing matched (so error) and if there's > 1 match then it's an error, too, because the user-provided name is ambiguous.

boardswarm-cli/src/main.rs
andrzejtp: done

boardswarm-cli/src/main.rs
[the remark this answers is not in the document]
sjoerdsimons: boardswarm doesn't generate `@<instance>` names though ;) Also no names aren't guaranteed to be unique the primary key is the id, the name is just a property

boardswarm-cli/src/main.rs
```diff
@@ -242,6 +242,76 @@ enum AuthCommand {
...
+    match arg {
+        ItemArg::Id(id) => Ok(id),
+        ItemArg::Name(name) => {
+            let mut items = client.list(item_type).await?;
+
+            let name_components: Vec<_> = name.split('@').collect();
+            if name_components.len() > 2 {
+                return Err(anyhow!("Too many @ in {} name", item_kind.1));
+            }
+
```
sjoerdsimons, line 271:
```suggestion
            let (name, instance) = name
                .rsplit_once('@')
                .map_or((name.as_str(), None), |(n, i)| (n, Some(i)));
```
no reason to create a vec and then do awkward indexing; you probably just want the name and an Option for the instance (as well it's optional)

boardswarm-cli/src/main.rs
```diff
@@ -242,6 +242,76 @@ enum AuthCommand {
...
+            /*
+             * first try finding a single exact match for the "name"
+             * part of the user-supplied "name[@instance]"
+             */
+            items.retain(|i| i.name == name_components[0]);
```
sjoerdsimons, line 276: this would retrain items regardless of whether the instance matches... And you don't end up checking for that later :)

If you want the semantic of the instance being optional you can use e.g. `i .name== name && (instance.is_none()  || i.instance == instance)` otherwise just `i.name == name && i.instance == instance`

boardswarm-cli/src/main.rs
```diff
@@ -242,6 +242,76 @@ enum AuthCommand {
...
+            match items.len() {
+                0 => Err(anyhow!("{} not found", item_kind.0)),
+                1 => Ok(items[0].id),
+                /* More matching entries found, try disambiguating by matching instance name */
```
sjoerdsimons, line 280: with the earlier suggestion you wouldn't need any of of this unless you want to prefer the local one on multiple matches, though i'm not sure if that's a wanted semantic

andrzejtp: I am afraid we _have to_ prefer local one on multiple matches because  there's no "@local". So if there are ttyS0, ttyS1 and ttyS0@remote, then the user saying "ttyS0" must mean the local one. And it's quite logical, because if the user wanted the remote one, they would say "ttyS0@remote".

sjoerdsimons (approved):
Add some extra small push on top; @andrzejtp would be great if you can let me know if you're happy with that additional change :)

PR #76, opened 2024-03-19
cli: Add help entry to "donwload-boot" rock command

andrzejtp:
download-boot is missing its help text.

### DIFF
--- a/boardswarm-cli/src/main.rs
+++ b/boardswarm-cli/src/main.rs
@@ -430,6 +430,7 @@ enum DeviceCommand {
 
 #[derive(Debug, Subcommand)]
 enum RockCommand {
+    /// Transfer a combined boot file containing images of type 0x471 and 0x472 to a rock device
     DownloadBoot {
         #[arg(short, long)]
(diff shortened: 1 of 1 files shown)


boardswarm-cli/src/main.rs
```diff
@@ -430,6 +430,7 @@ enum DeviceCommand {
 
 #[derive(Debug, Subcommand)]
 enum RockCommand {
+    /// Transfer concatenated bootrom images of type 0x471 and 0x472 to a rock device
```
sjoerdsimons, line 433: it's not a concatenation of bootrom images though; it's a combined boot file for which the exact rockchip jargon unfortunately escapes me

sjoerdsimons (changes requested):

boardswarm-cli/src/main.rs
andrzejtp: Let's then try: "Transfer a boot file to a rock device"

sjoerdsimons (approved):

Issue #37, opened 2024-02-26
cli: `write-bmap --wait` doesn't wait for volume to be available


obbardc:
I have a script which downloads the rockchip loader and then immediately tries to flash the eMMC:
```
boardswarm-cli rock rk3399 download-boot --wait rk3399_loader.bin
boardswarm-cli device rk3399 write-bmap --wait rockusb EMMC image.img.gz
```

This results in:
```
Waiting for volume..
0 Name: rk3399_ddr_800MHz_v
Done!... waiting 1ms
0 Name: rk3399_usbplug_v1.1
Done!... waiting 0ms
Error: status: Unavailable, message: "Target not found in volume", details: [], metadata: MetadataMap { headers: {} }
```

Sleeping for 5 seconds between commands makes it work, since the device's `EMMC` volume isn't instantly available as it is dynamically created by the loader, but I wonder if `write-bmap --wait` shouldn't wait (with a suitable timeout) already?

sjoerdsimons:
jargon (which may be confusing); Volume is a single storage volume or interface (e.g. `rockusb` named by the device config) which has one or more targets (in this case `EMMC`, which is dynamic depending on the provider). 

The problem you're seeing here is likely that the `rockusb` volume on will first come up in maskrom mode where it has the targets to write to sram/dram, which disappears once the usbplug is booted at which point it will come back with the `EMMC` target which is the volume you can write to.. However the `rockusb` maskrom volume will likely linger a bit till the plug resets the USB interface. Hence your write-bmap will see the volume it expects but not the target and errors out.

Two ways to fix this particular use-case; either download-boot should wait for the volume to disappear and/or write-bmap's wait should wait for the volume to come up with the right target. Fwiw this sometimes happens with dfu targets as well, but the dfu provider has a concept of reset which is trigger by doing a commit.

sjoerdsimons:
This was fixed some time ago :)

PR #77, opened 2024-03-20
Wait until unavailable in download boot

andrzejtp:
First attempt at solving https://github.com/boardswarm/boardswarm/issues/37

### DIFF
--- a/boardswarm-client/src/device.rs
+++ b/boardswarm-client/src/device.rs
@@ -67,6 +67,14 @@ impl DeviceVolume {
         }
     }
 
+    /// Wait until the volume becomes unavailable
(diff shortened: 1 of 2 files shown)


boardswarm-client/src/device.rs
```diff
@@ -67,6 +67,13 @@ impl DeviceVolume {
         }
     }
 
+    pub async fn wait_unavailable(&self) {
```
sjoerdsimons, line 70: Can you add a documentation string as well :)

sjoerdsimons (changes requested):

sjoerdsimons (approved):

PR #86, opened 2024-05-01
Add installation docs and tweak example files for ease of installation

mwelchuk:
Boardswarm currently lacks basic installation instructions, which would allow a potential user to efficiently evaluate the basic operation of boardswarm. Add a guide to the high level README.md covering the steps required to get an unencrypted setup installed and running with token based authentication.

Additionally tweak the example systemd unit and server configuration file inline with installation doumentation and make the installation process easier.

Note: This branch is based on the addition of the systemd unit and udev rules in #41.


README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+   $ sudo apt install rustc git protobuf-compiler libudev-dev
+   ```
+1. Clone the boardswarm repository:
+   ```
+   $ git clone https://github.com/boardswarm/boardswarm.git
+   ```
+1. We can now build boardswarm:
+   ```
+   $ cd boardswarm
+   $ cargo build
```
sjoerdsimons, line 37:
```suggestion
   $ cargo build --release
```
for normal usage you always want to build a release binary otherwise it's rather unoptimised

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+   $ sudo udevadm trigger
+   ```
+
+1. Copy example [`server.conf`](boardswarm/share/server.conf) to
+   `/etc/boardswarm/server.conf` on target server machine and modify as
+   required.
+
+### Server Configuration
+
+1. Install [pdudaemon](https://github.com/pdudaemon/pdudaemon) and configure it
```
sjoerdsimons, line 71: note that pdudaemon isn't mandatory; e.g. gpio actuators are available ;)  So may make sense to explain a bit more of why pdudaemon (wide support for programmable power supplies)

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+For each development board you will need a section under `devices`. There is an
+example one provided in the example server configuration which can be used as a
+template.
+
+#### USB Serial Consoles
+
+1. Determine the USB serial port used as the console on the development board.
+
+1. Use the following commmand to determine the required `udev.ID_SERIAL` and
+   `udev.ID_USB_INTERFACE_NUM`:
```
sjoerdsimons, line 107: needing `ID_USB_INTERFACE_NUM`; you only really have that for multi-port usb serial devices; also mind that ID_SERIAL is not always properly available on cheap usb->serial convertors and ofcourse not available for platform serial ports.

Fwiw another way of determining these propeties is using `boardswarm-cli consoles <console> properties`

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+   ```
+   $ udevadm info <tty_device> | grep -i 'ID_SERIAL=\|ID_USB_INTERFACE_NUM='
+   ```
+
+#### DFU Booting
+
+1. Power on the development board in DFU mode and run:
+   ```
+   $ sudo dfu-util -l
+   dfu-util 0.11
```
sjoerdsimons, line 117: unsure why we'd require dfu-util here ; boardswarm will automatically detect nodes as well

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+   Found DFU: [0451:6165] ver=0200, devnum=45, cfg=1, intf=0, path="3-3.4", alt=0, name="bootloader", serial="01.00.00.00"
+   ```
+
+1. Ensure that you have udev rules configuring the group ownership of usb
+   devices with the given USB vendor and product IDs (`0451:6165` in the above
+   example) to the boardswarm group.
+
+1. Use the following command to determine the required `ID_PATH`
+   ```
+   $ udevadm info /sys/bus/usb/devices/<path> | grep 'ID_PATH='
```
sjoerdsimons, line 134: similar you can get this from `boardswarm-cli list volumes` combined with `boardswarm-cli volume <volume> properties`

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+
+1. Determine the USB serial port used as the console on the development board.
+
+1. Use the following commmand to determine the required `udev.ID_SERIAL` and
+   `udev.ID_USB_INTERFACE_NUM`:
+   ```
+   $ udevadm info <tty_device> | grep -i 'ID_SERIAL=\|ID_USB_INTERFACE_NUM='
+   ```
+
+#### DFU Booting
```
sjoerdsimons, line 112: Might make sense  to create a document per protocol like this; as it makes sense to discuss each of them in details (we current support dfu and rockusb volumes)

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+1. Use the following command to determine the required `ID_PATH`
+   ```
+   $ udevadm info /sys/bus/usb/devices/<path> | grep 'ID_PATH='
+   ```
+   Where `<path>` is the path found by `dfu-util` (`3-3.4` in the above
+   example).
+
+### Run the Server
+
+Reload systemd configi, start the boardswarm server and configure it to run at
```
sjoerdsimons, line 141:
```suggestion
Reload systemd configuration, start the boardswarm server and configure it to run at
```

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+
+ * The devices configured on the server(s) can be listed:
+   ```
+   $ boardswarm-cli list devices
+   ```
+
+ * A board can be powered on and off:
+   ```
+   $ boardswarm-cli device <device_name> mode on
+   $ boardswarm-cli device <device_name> mode off
```
sjoerdsimons, line 180: caveate that the on and off names are conventions not guarantees

README.md
```diff
@@ -6,3 +6,188 @@ Distributed service to interact with development boards.
...
+   ```
+   $ rnbyc  -s '{"exp": 1735686000 }' -K private.jwks > token.jwt
+   ```
+
+1. Initialise the connection to target server:
+   ```
+   $ boardswarm-cli -i <server_name> auth init --uri http://<server>:6683 --token-file token.jwt
+   ```
+
+### Client Usage
```
sjoerdsimons, line 165: would it not be better to point to a dedicated client document that's part of the boardswarm-cli documentation ? rather then document the client in the server codebase

sjoerdsimons (changes requested):

obbardc (changes requested):

README.md
mwelchuk: Added

README.md
mwelchuk: Description improved to use boardswarm command and mention `ID_USB_INTERFACE_NUM` is only needed for multi-port devices.

README.md
mwelchuk: Replaced with boardswarm command.

README.md
mwelchuk: done

README.md
mwelchuk: done

README.md
mwelchuk: I have now mentioned this.

README.md
mwelchuk: Updated to mention that it's not mandatory and added more explanation

README.md
mwelchuk: I have worked the device configuration bits into the descriptions already in the server README.md as on reflection that seems a more suitable place to have this, given that there's already a good description of each of these there.

I haven't used rockusb yet...

README.md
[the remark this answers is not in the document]
mwelchuk: Does it make sense to document that here? I've not documented installing rust either...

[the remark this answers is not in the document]
obbardc: Perhaps you could say about installing cargo.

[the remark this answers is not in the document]
mwelchuk: Hang on - I *am* already documenting installing rust and the dependencies for building boardswarm and this automatically pulls in cargo, so this is already documented...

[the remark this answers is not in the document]
mwelchuk: (At least on Debian)

README.md
```diff
@@ -37,3 +37,134 @@ docker run -it --rm \
...
+
+## Manual Installation
+
+These instructions are for configuring a simple unencrypted setup with static
+JWKS based authentication.
+
+### Building
+
+Boardswarm needs rustc version 1.70 or newer, this rules out a number of stable
+distributions for now (the kind of thing you may want to run your
```
sjoerdsimons, line 49: It only rules out the rustc available in the distro itself though (which sure fair), but lots of folks end up installing rust with rustup in which case it works well (or do the build in a container as you don't want dev dependencies on your server)

This paragraph is a bit convoluted because it assumes building with a rustc provided by eg.. debian though that's not entirely clear :)..

Fwiw to avoid people have to install rust etc i've been playing with creating pre-build binaries (next to the containers) (see the dist branch), unfortunately github makes that a tad awkward as you can't easily publish artifacts on a brnach without making a release

mwelchuk: As it turns out, the path I'd suggested it seems can also break if the libc on the building machine and the one on the target machine are sufficiently different:
```
boardswarm[54132]: /usr/local/bin/boardswarm: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.38' not found (required by /usr/local/bin/boardswarm)
```

mwelchuk: Significantly reworked building instructions.

boardswarm/README.md
```diff
@@ -318,6 +343,46 @@ devices:
           udev.ID_PATH: "pci-0000:00:14.0-usb-0:12.3"
 ```
 
+#### DFU Volumes
```
sjoerdsimons, line 346: This section will apply to any kind of volume (dfu, rockusb) not just dfu

boardswarm-cli/README.md
```diff
@@ -34,3 +45,12 @@ The UI has the follow keyboard shortcuts:
 * ^a k: Scroll up
 * ^a j: Scroll down
 * ^a 0: Reset scrolling state
+
+## Boardswarm device control
+
+Basic board control can be performed without opening the tui, for example
+a board can be powered on and off:
```
sjoerdsimons, line 52: Mind that's only the case if the mode are configure like on/off  ofcourse; those names are just conventions

README.md
```diff
@@ -37,3 +37,175 @@ docker run -it --rm \
...
+$ sudo systectl enable boardswarm.service
+$ sudo systemctl start boardswarm.service
+```
+
+### Client Installation
+
+1. Copy the `boardswarm-cli` binary to `/usr/local/bin/` on the client
+   development machine.
+
+1. Create a user token (expiry 01-01-2025) using the previously generated
```
sjoerdsimons, line 197: This is duplicated from the boardswarm `static jwks auth` section

README.md
```diff
@@ -37,3 +37,175 @@ docker run -it --rm \
   -v $(pwd)/boardswarm/share/server.conf:/etc/boardswarm/server.conf \
   ghcr.io/boardswarm/boardswarm:main
 ```
+
+## Manual Installation
+
+These instructions are for configuring a simple unencrypted setup with static
+JWKS based authentication.
```
sjoerdsimons, line 44: Better to move in the boardswarm readme as that will also be distributed with the crate

README.md
```diff
@@ -37,3 +37,175 @@ docker run -it --rm \
...
+   and complain if the expected version isn't found, making this solution
+   unreliable at times.) For instance, assuming a Debian system:
+   ```
+   $ sudo apt install rustc
+   ```
+
+ - If installing software outside of the systems package management isn't for
+   you and building elsewhere isn't an option, then creating a container from
+   which to build boardswarm might be a solution. A simple Dockerfile to just
+   build boardswarm is as follows:
```
sjoerdsimons, line 68: I don't like having that untestable in the documentation; It just means it'll bitrot

README.md
```diff
@@ -37,3 +37,175 @@ docker run -it --rm \
...
+$ cargo build --release
+```
+Or, if wanting to use a container for this, run the following command:
+```
+podman build --volume=$(pwd):/usr/src/boardswarm .
+```
+
+### Server Install
+
+The "server" is the machine that will have the development boards attached to
```
sjoerdsimons, line 114: no need to have board attached per se ; it can purely act as a router as well (connecting to other boardswarms)

README.md
```diff
@@ -37,3 +37,175 @@ docker run -it --rm \
...
+   ```
+
+1. We are going to use Static JWKS based authentication, as documented in the
+   [server README](boardswarm/README.md#static-jwks-based-authentication).
+   Install `rnbyc` so we can create the required tokens:
+   ```
+   $ sudo apt install rnbyc
+   ```
+
+1. Generate an EdDSA key:
```
sjoerdsimons, line 163: rather not have these instructions in two places; they're bound to get out of sync

README.md
mwelchuk: I'm not sure how to best resolve this. I was unable to build boardswarm on a Debian Stable machine that I'd got setup for my lab control. Building on Debian Testing worked, but then wouldn't run on the Stable machine due to different versions of libc being used. I doubt I'm the only one that prefers not to stuff a load of `curl | sh` installs on systems, especially those that I hope to keep running without complete reinstalls for extended periods. The existing Dockerfiles don't provide a means to access the binaries outside of Docker once they've been built.

Would having this as a Dockerfile in the root directory of the project be an option you'd consider?

README.md
mwelchuk: Would splitting out the server portion and client portion into the relevant READMEs and leaving just some pointers here to the individual parts make sense?

boardswarm/README.md
mwelchuk: Generalised.

README.md
mwelchuk: Duplication removed.

README.md
mwelchuk: Added note about use as router.

README.md
sjoerdsimons: Yeah that would be great; Generally the central README should be mostly pointers with most content in the respective READMEs so that if people enter through e.g. crates.io or anything else with the parts seperated everything relevant is there

README.md
sjoerdsimons: We have already two docker files from which you seem to have copied this from; if you split the builder stage into a first stage that creates the builder and second one that does the build then you can easily build just the build container from that if you prefer.

Also while you're right, especially at this point of development, i don't think the type of audience that might give boardswarm a go *and* wnats to build it themselves should have to be thought how top installer Rust per se :).

for those that don't want or cant themselves easily, but do run bookworm (or compatibl repos); that's why i'm doing https://github.com/sjoerdsimons/boardswarm/tree/dist to allow downloading regularily build tarballs (which i'd also want to use for my own deployments)

README.md
mwelchuk: Hmm, been trying to do this, but basically to make the docs then flow, it goes from adding a "quick start guide" as I'd envisioned. to "marginal improvements to relatively lengthy documentation". Which wasn't exactly the intention...

README.md
```diff
@@ -37,3 +37,166 @@ docker run -it --rm \
...
+
+### Client Installation
+
+1. Copy the `boardswarm-cli` binary to `/usr/local/bin/` on the client
+   development machine.
+
+1. Create a user token (expiry 01-01-2025) using the previously generated
+   `private.jwks` and then copy it to the client machine:
+   ```
+   $ rnbyc  -s '{"exp": 1735686000 }' -K private.jwks > token.jwt
```
obbardc, line 191: Just FYI, this had already expired. Perhaps the command could be updated to expire e.g a year from when its ran?

Issue #81, opened 2024-04-04
pdudaemon actuator changes state even though it failed


obbardc:
pdudaemon seems to always return true to a request, even if a port fails to switch: https://github.com/pdudaemon/pdudaemon/issues/156

This can cause a device's mode state to change even though the hardware hasn't actually changed.

This task is about making sure this is fixed in boardswarm and https://github.com/sjoerdsimons/pdudaemon-client once it has been fixed in pdudaemon.

obbardc:
So I managed to hack together handling errors in pdudaemon: https://github.com/pdudaemon/pdudaemon/pull/158


After a quick look in boardswarm we don't actually handle pdudaemon errors in boardswarm and just panic instead:

```
thread 'tokio-runtime-worker' panicked at boardswarm/src/pdudaemon.rs:108:72:
called `Result::unwrap()` on an `Err` value: ReqwestError(reqwest::Error { kind: Status(500), url: Url { scheme: "http", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("localhost")), port: Some(16421), path: "/power/control/off", query: Some("hostname=pdu&port=4"), fragment: None } })
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Need to not panic on error on the two calls in `on` and `off` at https://github.com/boardswarm/boardswarm/blob/main/boardswarm/src/pdudaemon.rs#L106

PR #89, opened 2024-05-09
Graciously handle pdudaemon errors

obbardc:
Currently boardswarm panics if pdudaemon returns an error. Make it a little more gracious and bubble up an error instead.

This is not currently a problem, since pdudaemon will not return errors until https://github.com/pdudaemon/pdudaemon/pull/158 is merged.

Fixes: boardswarm/boardswarm#81

### DIFF
--- a/boardswarm/src/pdudaemon.rs
+++ b/boardswarm/src/pdudaemon.rs
@@ -6,7 +6,7 @@ use tracing::instrument;
 
 use crate::{
     registry::{self, Properties},
-    Server,
+    ActuatorError, Server,
 };
 
 pub const PROVIDER: &str = "pdudaemon";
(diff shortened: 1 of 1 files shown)


boardswarm/src/pdudaemon.rs
```diff
@@ -97,17 +97,17 @@ impl crate::Actuator for PduDaemonActuator {
...
         let parameters = ModeParameters::deserialize(parameters).unwrap();
         match parameters.mode.as_str() {
-            "on" => self.daemon.on(&self.hostname, &self.port).await.unwrap(),
-            "off" => self.daemon.off(&self.hostname, &self.port).await.unwrap(),
+            "on" => self.daemon.on(&self.hostname, &self.port).await,
+            "off" => self.daemon.off(&self.hostname, &self.port).await,
             _ => todo!(),
         }
-        Ok(())
+        .map_err(|_e| ActuatorError {})
```
obbardc, line 111: Not sure if this is the right thing to do, possibly we could return the actual error from the daemon. But it currently works!

sjoerdsimons: Ideally the errors from actuators should be more descriptive but this still is a net improvement :)

sjoerdsimons (changes requested):
Can you fix the typos in the commit message; otherwise seems fine to me :)

boardswarm/src/pdudaemon.rs
obbardc: I created https://github.com/boardswarm/boardswarm/issues/92 to handle that. I doubt I will have the time to do that here :-).

obbardc:
@sjoerdsimons I think I made the commit message better.

sjoerdsimons:
looks like the commit message still says "pdudameon" instead of "pdudaemon" :)

obbardc:
> looks like the commit message still says "pdudameon" instead of "pdudaemon" :)

@sjoerdsimons I completely missed that it was spelt incorrectly, even on a second time, can I kindly ask if you mention "typos" that you specifically call out what they are as that would stop one extra round of back-and-forth? I've fixed it now anyway :-)

sjoerdsimons (approved):

Issue #14, opened 2023-12-12
boardswarm-cli ui mode could be more obvious after changing device mode


obbardc:
When scrolling in UI mode (with `^a k`, `^a j`), a small green area is shown with the line number.

I wonder if it'd be good to show a general status bar, with some user feedback. An immediate one I can think of is showing state e.g. `Changed device mode to X` when changing device state.

obbardc:
An option could be to rework the tui using something like https://github.com/ratatui-org/ratatui

sjoerdsimons:
> An option could be to rework the tui using something like https://github.com/ratatui-org/ratatui

Fwiw the UI already uses tui from which ratatui was forked ;) we for sure should move to ratatui indeed but that doesn't impact your other requests

obbardc:
Yeah it wasn't clear to me that ratatui was a continuation project rather than a full rewrite until you mentioned it :-).
I moved that forward in #16 anyway, but for sure adding a status bar and other UI bits isn't blocked on that.

Issue #13, opened 2023-12-12
boardswarm-cli ui mode hangs when client panics


obbardc:
in UI mode, I tried to switch a device on which was already on. The UI then panic'd with the below message and then the shell didn't accept any user input.

```
thread 'tokio-runtime-worker' panicked at boardswarm-cli/src/ui.rs:225:60:                                                                                                                                                                                                                                                              
                                                                                              called `Result::unwrap()` on an `Err` value: Status { code: FailedPrecondition, message: "Not in the right mode to switch", metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Tue, 12 Dec 2023 10:08:33 GMT", "content-length": "0"} }, source: None }                                                                                                                                                                                                                                                                                                                                       
                     note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace                                                                                                                                                                                                                                                          
```

obbardc:
@andrzejtp did you manage to look at this?

andrzejtp:
A minimal fix (workaround?) is to simply ignore the Err here: https://github.com/boardswarm/boardswarm/blob/aa3a7bf91d47d9e5706afa677572314b9c99d8c1/boardswarm-cli/src/ui.rs#L226

e.g. like this:

```rust
                            let _ = device.change_mode("on").await;
                            None
```

I am unable to tell if this is appropriate or not, though. In other words, I am not sure if by doing this fix I am also ignoring other conditions which _should_ result in a panic.

obbardc:
yeah, I think the first step should just be not crashing the whole UI when we fail to set a mode from the keypresses and instead warn cleanly :-)

adalessandro:
Pushed https://github.com/boardswarm/boardswarm/pull/171 
* Added status bar.
* Panic handled, error message shown in status bar.

PR #90, opened 2024-05-14
Show device mode

andrzejtp:
This is an attempt to address https://github.com/boardswarm/boardswarm/issues/14

There's an "Initial" state of the device. Maybe we should instead somehow query the device for its initial state. 

Perhaps we could also want to show device's `id` and/or `boardswarm.name`.


boardswarm-cli/src/ui.rs
```diff
@@ -223,23 +259,35 @@ pub async fn run_ui(
                 async move {
                     match i {
                         Input::PowerOn => {
-                            device.change_mode("on").await.unwrap();
+                            let _ = device.change_mode("on").await;
```
sjoerdsimons, line 262: while just unwrapping isn't great; hiding the failure isn't good either :)

boardswarm-cli/src/ui_term.rs
```diff
@@ -5,13 +5,31 @@ use ratatui::{
     widgets::Widget,
 };
 
+#[derive(Clone, Copy, Debug)]
+pub enum TerminalState {
```
sjoerdsimons, line 9: These are not states; just device modes and device modes are not enums they're aribitrary strings (on/off are just conventions); Reset is not a state either and neither is initial.

boardswarm-cli/src/ui_term.rs
```diff
@@ -56,6 +74,20 @@ impl Widget for UiTerm<'_> {
...
+            let width = str.len() as u16;
+            let style = Style::reset().fg(Color::Black);
+            let style = match self.terminal_state {
+                TerminalState::PowerOff => style.bg(Color::LightRed),
+                TerminalState::PowerOn => style.bg(Color::LightGreen),
+                TerminalState::PowerReset => style.bg(Color::LightYellow),
+                TerminalState::Initial => style.bg(Color::Gray),
+            };
+            let span = Span::styled(str, style);
+            buf.set_span(area.width - width, area.height - 1, &span, width);
```
sjoerdsimons, line 89: this probably should not be in the uiterm widget but instead in e.g. a status bar widget

boardswarm-cli/src/ui.rs
```diff
@@ -206,7 +238,11 @@ pub async fn run_ui(
                         Input::Up => { terminal.scroll_up(); },
                         Input::Down => { terminal.scroll_down(); },
                         Input::ScrollReset => { terminal.scroll_reset(); },
-                        _ => (),
+                        Input::PowerOn => { terminal.set_power_on(); },
+                        Input::PowerOff => { terminal.set_power_off(); },
+                        Input::PowerReset => { terminal.set_power_reset(); },
+                        Input::ShowState => { terminal.show_state(); },
```
sjoerdsimons, line 244: Really the device mode should be coming from the actual device mode; not what a ui has sent ..  You can listen to the device-info stream ; See how e.g. the implementation of `boardswarm-cli device X info --follow`

sjoerdsimons (changes requested):

boardswarm-cli/src/ui.rs
andrzejtp: Would the right place to listen to device info stream be https://github.com/boardswarm/boardswarm/blob/7c04a7c8d41ba0fbfdb2892926157c7c93a3b1f1/boardswarm-cli/src/ui.rs#L196?

sjoerdsimons: I'd probably do it in a seperate task and only message the main thread when there is a mode update specificially such that it can trigger a next frame.

boardswarm-cli/src/ui_term.rs
andrzejtp: OTOH if I want to display different modes with different attributes (e.g. on would be black on green, while off would be black on red), then I need to start interpreting the string, which seems not the greatest option. What's more, future localization might mean that actual strings passed are unknown to the programmer.

andrzejtp:
Added layout and a separate widget for device mode

obbardc (changes requested):

boardswarm-cli/README.md
```diff
@@ -31,6 +31,7 @@ The UI has the follow keyboard shortcuts:
 * ^a o: Change the device to mode "on"
 * ^a f: Change the device to mode "off"
 * ^a r: Reset the device's power (same as changing mode to "off" then "on")
+* ^a i: Show device's current mode
```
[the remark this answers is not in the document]
andrzejtp: Have you tried actually running the code?

The idea is that the mode _is_ displayed automatically on any change. But then after the user starts typing the mode becomes _not displayed_ in order to not clutter the terminal. Should the user want to recall what the current mode is, they can issue ctrl-a i - the mode is then shown and is displayed until the user starts typing again.

boardswarm-cli/src/ui.rs
```diff
@@ -206,14 +269,23 @@ pub async fn run_ui(
...
                         Input::Down => { terminal.scroll_down(); },
                         Input::ScrollReset => { terminal.scroll_reset(); },
-                        _ => (),
+                        Input::ShowMode => { terminal.show_mode(); },
+                        _ => { terminal.hide_mode(); },
+                    }
+                }
+                Some(new_state) = state_rx.recv() => {
+                    if new_state == "on" {
+                        terminal.set_power_on("PowerOn".to_string());
```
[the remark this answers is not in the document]
andrzejtp: I think it makes sense to decouple the mode from definition of how mode is displayed. By that I mean that if e.g. "off" is displayed as black on red, maybe other kinds of modes will also want to be displayed black on red, so I called this kind of display "NonFunctional", as opposed to "Functional" which is black on green. That is why the modes are not relayed 1:1 to the ui. On top of that, the string needs to be parsed to know how to display it. Not great maybe. And then if localization (i.e. displaying in national languages rather than English) is added, then the strings passed will not be known up front to the programmer, so the programmer will effectively be unable to parse (in a sane way) even if they wanted to.

For now this question boils down to "where should the mode string be parsed in order to decide how the mode is displayed?". Any preference/rationale for that?

boardswarm-cli/src/ui.rs
```diff
@@ -235,17 +307,38 @@ pub async fn run_ui(
...
+                        }
                         Input::Up | Input::Down | Input::ScrollReset => {
                             input_tx.send(i).await.unwrap();
                             None
                         }
-                        Input::Bytes(data) => Some(data),
+                        Input::Bytes(data) => {
+                            // send an empty "Bytes" to notify that some
+                            // ordinary characters have been entered
+                            input_tx.send(Input::Bytes(Bytes::new())).await.unwrap();
```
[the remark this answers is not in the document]
andrzejtp: This is directly related to mode ceasing being displayed as soon as the user starts typing.

boardswarm-client/src/device.rs
```diff
@@ -232,6 +232,14 @@ impl Device {
         Ok(())
     }
 
+    pub async fn device_info(
```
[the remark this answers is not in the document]
andrzejtp: :+1:

boardswarm-cli/src/ui.rs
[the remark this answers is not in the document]
obbardc: oh that really wasn't clear :-) perhaps you could add a comment about what is going on ?

boardswarm-cli/README.md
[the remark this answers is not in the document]
obbardc: No this is purely a review at this point ;-)

i wonder why it makes sense to hide it ? possibly we could always keep a row at the top/bottom of the screen for the boardswarm ui (with the console taking up -2 rows)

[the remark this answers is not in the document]
andrzejtp: For now the terminal is fixed at 80x24. Permanently stealing one or even two lines from that seems a crime to me.

andrzejtp:
@obbardc This time I have simplified the code a bit. The status is displayed always, which uses the line in the bottom of the display. That's not ideal, because we're left with just 23 lines, but OTOH relaxing 80x24 is probably an orthogonal problem.

boardswarm-cli/src/ui.rs
```diff
@@ -209,11 +259,19 @@ pub async fn run_ui(
                         _ => (),
                     }
                 }
+                Some(new_state) = state_rx.recv() => {
```
obbardc, line 262: how about other states ? can we just pass them straight through to the UI ?

a match statememnt may be cleaner here

obbardc (changes requested):

andrzejtp:
@obbardc In this iteration on top of addressing your comment about matching the state string, I've addressed https://github.com/boardswarm/boardswarm/issues/13 in such a way that we don't hang but instead the message is relayed to the message area which I've split from the display bottom line.

andrzejtp:
@obbardc I mistakenly pushed an old branch and corrected it by pushing the latest changes. This time https://github.com/boardswarm/boardswarm/issues/13 is also addressed by relaying the message to the user instead of hanging at unwrap().

boardswarm-client/src/device.rs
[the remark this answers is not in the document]
sjoerdsimons: the Device already monitors the info stream; so with this api you're now receiving two info streams;  The Device struct was really meant for high-level api's, so it's a bit odd to directly expose the low-level protocol like that again

boardswarm-cli/src/ui.rs
```diff
@@ -203,17 +289,30 @@ pub async fn run_ui(
...
+                                _ => (),
+                            },
+                        InputMsg::Error(e) => { terminal.set_message(e); },
+                    }
+                }
+                Some(new_state) = state_rx.recv() => {
+                    let state: String = new_state;
+                    match state.as_str() {
+                        "on" => terminal.set_functioning("ON".to_string()),
+                        "off" => terminal.set_non_functioning("OFF".to_string()),
```
sjoerdsimons, line 306: these are devie modes not terminal modes though; functioning/non-functioning is weird jargon as welll ;  why not just on/off/other(string) ?

boardswarm-cli/src/ui.rs
```diff
@@ -50,21 +70,79 @@ impl Terminal {
...
         let term = ui_term::UiTerm::new(screen);
+        let terminal_mode = self.terminal_mode.clone();
+        let message = self.message.clone();
         self.tui
             .draw(|f| {
                 let size = f.size();
-                let term_area = Rect::new(0, 0, 80.min(size.width), 24.min(size.height));
-                f.render_widget(term, term_area);
+                let term_area = Rect::new(0, 0, COLUMNS.min(size.width), ROWS.min(size.height));
+                let area_chunks = Layout::default()
```
sjoerdsimons, line 105: fwiw you can use `.areas` and destructure instead of having to do array indexes

boardswarm-cli/src/ui.rs
```diff
@@ -50,21 +70,79 @@ impl Terminal {
...
+                    .direction(Direction::Vertical)
+                    .constraints([Constraint::Min(1), Constraint::Length(1)])
```
sjoerdsimons, line 107: I assume this means that if the terminal size is actually 80x24 you'll only see 23 lines? Would it make sense to add a min constraint of 24 for the term so we prioritize over showing that fully *over* showing a status bar?

boardswarm-cli/src/ui.rs
```diff
@@ -246,6 +350,18 @@ pub async fn run_ui(
             .await
     });
 
+    let _device_state_watcher = tokio::spawn(async move {
+        let mut d = dev_clone.device_info().await.unwrap();
+        let mut current_mode = None;
```
sjoerdsimons, line 355: The device mode already knows the current mode as it will wait for a first info on creation.

boardswarm-client/src/device.rs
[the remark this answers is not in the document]
sjoerdsimons: fwiw looking overall; it may make sense to push watching the device mode into the Device itself; either with a one-shot api or by exposing a stream of mode-changes. Should probably also add an api to get the current mode

sjoerdsimons (changes requested):

boardswarm-cli/src/ui.rs
andrzejtp: I'm okay with renaming `TerminalMode` to `DeviceMode`.

Functioning/NonFunctioning: Perhaps there will be states which want to be displayed the same way "off" or "on" is displayed. Let's imagine there will appear a state called "relay" which wants to be displayed black on green. Then the match arm becomes:

```
"relay" => terminal.set_functioning("RELAY".to_string()),
```

Otherwise it will have to be:

```
"relay" => terminal.set_on("RELAY".to_string()),
```

boardswarm-cli/src/ui.rs
andrzejtp: Yes, in the current shape of the series I do steal the lowest line for the status bar.

You guys settle on what you want first https://github.com/boardswarm/boardswarm/pull/90#discussion_r1622021635

Actually I imagine that you maybe _don't_ know and this thread is exactly about _learning_ what we want and I'm okay with that. Only I wouldn't like to go in circles too many times. ;P

It seems to me that relaxing the 80x24 is a bit of an orthogonal problem. IOW trying to conditionally display the status bar (sometimes display, sometimes not) is effectively trying to work around not being able to use other terminal sizes. Then there's a question of _when_ to display and when not. Then the user might want to recall what the current mode is. It's already been tried in this thread. Displaying the status bar all the time does simplify the code.

sjoerdsimons: I don't need to settle on anything really :) Displaying the console is the obvious primary function of the UI everything else is secondary (hence that's all it does at the moment); We shouldn't ever force the terminal area to be smaller then what it wants to be as that will likely hide output lines; e.g. if the last line is hidden your shell input will disappear quite quickly; if the top line is hidden you're missing an important part of e.g. `top`.

Not being able to set the size is orthogonal to this and primarily it's not really a boardswarm client call but depends on what the remote console side thinks it's rendering to, so we can't just decide "relax" and change size because we want to fit a statusbar :) Imho this should go as far as the UI displaying an error if the output size is smaller then its terminal size, because, really, we should never hide things in the terminal as that quickly ends up confusing. 

For 80x24, that's still the default when opening most terminal emulators these days even if it's somewhat historical. And similar a serial tty on linux/u-boot serial output etc all also start at 80x24, hence the hardcoding in the UI at the moment. But that also fundamentally means we should support displaying 80x24  on a 80x24, hence dropping out the status bar in those cases to not be annoying in that situation :)

boardswarm-cli/README.md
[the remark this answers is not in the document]
sjoerdsimons: We should indeed never steal lines; A permanent status bar is a nice to have but it should never be prioritised over the terminal

boardswarm-cli/src/ui.rs
sjoerdsimons: can you please not yell in strings ;) Yes there might be cases we may want to do differently (or maybe we could stick a hint in the protocol for broad categories), but regardless we can simply add more "known" items in the enum. 

Fwiw it would also be nicer to implement From<&str> for your mode enum

boardswarm-cli/README.md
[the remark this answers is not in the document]
adalessandro: Is there a particular reason why the area is fixed to 80x24?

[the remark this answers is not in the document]
sjoerdsimons: @adalessandro  see https://github.com/boardswarm/boardswarm/pull/153#pullrequestreview-2560369986

PR #142, opened 2024-12-09
Mediatek brom and fastboot volume support

sjoerdsimons:


obbardc:
nice, fastboot support !

Issue #96, opened 2024-05-22
ui mode causes terminal corruption when server returns an error


obbardc:
Client starts UI mode for a device:
```
$ boardswarm-cli -i instance ui device
```

Server has some (unrelated to this ticket) issue opening serial port:
```
boardswarm[19269]: thread 'tokio-runtime-worker' panicked at boardswarm/src/serial/mod.rs:122:27:
boardswarm[19269]: called `Result::unwrap()` on an `Err` value: Permission denied
```

this causes an error in the client, the UI mode exits incorrectly which causes terminal corruption which needs to be reset using `reset`:
```
Error: status: Cancelled, message: "h2 protocol error: http2 error: stream error received: stream no longer needed", details: [], metadata: MetadataMap { headers: {} }                                                                                                                     

Caused by:
0: transport error
1: http2 error: stream error received: stream no longer needed
2: stream error received: stream no longer needed
```

Possibly similar to #13 but thought this deserved its own ticket. We should 1) make sure this doesn't hang. 2) see if we can print a nicer error on the client-side?

PR #151, opened 2024-12-30
Wip/laeyraud/fix UI mode hang on panics

laeyraud:
This merge request aims to prevent from client side the panics in UI mode and to restore the terminal properly when an error occurs. It should fix both #13 and #96 issues.

Two behaviours are also modified:
* If a non-fatal error occurs when trying to connect to a device, the user displayed message is now more explicit  and shorter.
* If an action fails to process, the device console is used to display the message as a new line.


boardswarm-cli/src/ui.rs
```diff
@@ -210,36 +206,64 @@ pub async fn run_ui(
...
+                                input_tx.send(Input::UIMessage(err_msg)).await.unwrap();
+                            }
                             None
                         }
                         Input::PowerOff => {
-                            device.change_mode("off").await.unwrap();
+                            // Perform power off action
+                            if let Err(err) = device.change_mode("off").await {
+                                let err_msg: String =
+                                    format!("Change mode 'off' action failed: {}\r\n", err.code());
```
sjoerdsimons, line 221: seems like you're copying this error handling 4 times now; probably make sense to stick that in a helper function

boardswarm-cli/src/ui.rs
```diff
@@ -210,36 +206,64 @@ pub async fn run_ui(
...
+                            if let Err(err) = device.change_mode("on").await {
+                                let err_msg: String =
+                                    format!("Change mode 'on' action failed: {}\r\n", err.code());
+                                let _ = input_tx.send(Input::UIMessage(err_msg)).await;
+                            }
                             None
                         }
                         Input::Up | Input::Down | Input::ScrollReset => {
-                            input_tx.send(i).await.unwrap();
+                            let _ = input_tx.send(i).await;
```
sjoerdsimons, line 242: not a blocker; but if we can send on this channels something is very very wrong ofcourse

boardswarm-cli/src/ui.rs
```diff
@@ -193,6 +188,7 @@ pub async fn run_ui(
                         Input::Up => { terminal.scroll_up(); },
                         Input::Down => { terminal.scroll_down(); },
                         Input::ScrollReset => { terminal.scroll_reset(); },
+                        Input::UIMessage(msg) => {terminal.process(msg.as_bytes())}
```
sjoerdsimons, line 191: This will ofcourse just dump in the error message as if it's e.g. output from the remote terminal; While for sure it's an improvement on the current state - it's not great. Ideally we'd throw the errors on top of the terminal in a new window instead.

boardswarm-cli/src/ui.rs
laeyraud: Sure, I'll create a function to handle both change_mode call and the error handling

boardswarm-cli/src/ui.rs
adalessandro: Indeed, makes sense to panic here, as it means something is really broken.

boardswarm-cli/src/ui.rs
laeyraud: I'll drop this commit when you finish #171 then I'll rebase this MR on it.

obbardc: Is there any plans to update this? Ran into the problem today and it reminded me to check :-)

PR #152, opened 2024-12-30
Fix panic on serial port opening failure

laeyraud:
This commit prevents a panic if a serial port cannot be opened (eg permission issues).

### DIFF
--- a/boardswarm/src/main.rs
+++ b/boardswarm/src/main.rs
@@ -52,8 +52,19 @@ trait Actuator: std::fmt::Debug + Send + Sync {
 
 #[derive(Error, Debug)]
 pub enum ConsoleError {
+    #[error("Unavailable: {0}")]
+    Unavailable(String),
     #[error("Console was closed")]
     Closed,
+}
+
+impl From<ConsoleError> for tonic::Status {
+    fn from(e: ConsoleError) -> Self {
+        match e {
+            ConsoleError::Closed => tonic::Status::aborted(e.to_string()),
+            ConsoleError::Unavailable(msg) => tonic::Status::unavailable(msg),
+        }
+    }
 }
 
 #[async_trait::async_trait]
@@ -74,12 +85,12 @@ type ConsoleOutputStream =
 
 #[async_trait::async_trait]
 trait ConsoleExt: Console {
-    async fn output_stream(&self) -> ConsoleOutputStream {
-        Box::pin(self.output().await.unwrap().map(|data| {
+    async fn output_stream(&self) -> Result<ConsoleOutputStream, ConsoleError> {
+        Ok(Box::pin(self.output().await?.map(|data| {
             Ok(boardswarm_protocol::ConsoleOutput {
                 data: data.unwrap(),
             })
-        }))
+        })))
     }
 }
 
@@ -745,7 +756,8 @@ impl boardswarm_protocol::boardswarm_server::Boardswarm for Server {
     ) -> Result<tonic::Response<Self::ConsoleStreamOutputStream>, tonic::Status> {
         let inner = request.into_inner();
         if let Some(console) = self.get_console(inner.console) {
-            Ok(tonic::Response::new(console.output_stream().await))
+            let stream = console.output_stream().await?;
+            Ok(tonic::Response::new(stream))
         } else {
             Err(tonic::Status::invalid_argument("Can't find output console"))
         }
--- a/boardswarm/src/serial/mod.rs
+++ b/boardswarm/src/serial/mod.rs
@@ -137,24 +137,34 @@ impl SerialPort {
         Ok(())
     }
 
-    async fn get_writer(&self) -> Arc<AsyncMutex<WriteHalf<SerialStream>>> {
+    #[instrument(skip_all, err)]
+    async fn get_writer(&self) -> Result<Arc<AsyncMutex<WriteHalf<SerialStream>>>, ConsoleError> {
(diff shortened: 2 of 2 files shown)


boardswarm/src/serial/mod.rs
```diff
@@ -137,24 +137,42 @@ impl SerialPort {
...
-    async fn get_writer(&self) -> Arc<AsyncMutex<WriteHalf<SerialStream>>> {
+    async fn get_writer(&self) -> Result<Arc<AsyncMutex<WriteHalf<SerialStream>>>, ConsoleError> {
         if let Some(open) = &*self.open.lock().await {
-            return open.write.clone();
+            return Ok(open.write.clone());
         }
-        self.open().await.unwrap();
+
+        if let Err(e) = self.open().await {
+            let fmt_err = ConsoleError::InternalError(format!(
```
sjoerdsimons, line 146: This isn't really an internal error; an internal error would mean something is very broken... internally  (there are probably other parts of the codebase that abuse it a bit).

while serial opens failing are often either due to the device having disappeared in the mean time (race fun), because the permissions are incorrect or something is actually wrong with the serial console.. Which could be mapped to respectively `not_found`, `unavailable` and probably `unknown` ? Though as a first step i wouldn't mind having it all map to one of those (either unavailable or unknown).

boardswarm/src/serial/mod.rs
```diff
@@ -137,24 +137,42 @@ impl SerialPort {
         Ok(())
     }
 
-    async fn get_writer(&self) -> Arc<AsyncMutex<WriteHalf<SerialStream>>> {
+    async fn get_writer(&self) -> Result<Arc<AsyncMutex<WriteHalf<SerialStream>>>, ConsoleError> {
```
sjoerdsimons, line 140:
```suggestion
   #[instrument(skip_all, err)]
    async fn get_writer(&self) -> Result<Arc<AsyncMutex<WriteHalf<SerialStream>>>, ConsoleError> {
```
This will automatically cause an error log to be triggered when the function returns an error

boardswarm/src/serial/mod.rs
```diff
@@ -137,24 +137,42 @@ impl SerialPort {
...
         }
-        self.open().await.unwrap();
+
+        if let Err(e) = self.open().await {
+            let fmt_err = ConsoleError::InternalError(format!(
+                "Failed to open '{}' serial: {}",
+                self.path, e
+            ));
+            error!("{}", fmt_err.to_string());
+            return Err(fmt_err);
```
sjoerdsimons, line 151: `map_err` would likely be more elegant here

boardswarm/src/serial/mod.rs
```diff
@@ -137,24 +137,42 @@ impl SerialPort {
...
+
         if let Some(open) = &*self.open.lock().await {
-            return open.write.clone();
+            return Ok(open.write.clone());
         }
         unreachable!();
     }
 
-    async fn get_reader(&self) -> broadcast::Receiver<Bytes> {
+    async fn get_reader(&self) -> Result<broadcast::Receiver<Bytes>, ConsoleError> {
```
sjoerdsimons, line 160: same comments as get_writer :)

sjoerdsimons (changes requested):

boardswarm/src/serial/mod.rs
laeyraud: You're right, unavailable would be better in this case.
I'll replace this new InternalError value by Unavailable in ConsoleError enum and use it instead in the code.

boardswarm/src/serial/mod.rs
laeyraud: I did not know this macro, I'll try it as it would make the new error handling more compact

boardswarm/src/serial/mod.rs
laeyraud: You're right.
With the macro automatically printing the error, it allows to compact the open error handling in both get_reader and get_writer.

boardswarm/src/serial/mod.rs
laeyraud: I'll apply the same modifications in the function.

laeyraud:
I've pushed the requested fixes and rebased the branch with latest commits from main at the same time

boardswarm/src/main.rs
```diff
@@ -745,7 +747,11 @@ impl boardswarm_protocol::boardswarm_server::Boardswarm for Server {
     ) -> Result<tonic::Response<Self::ConsoleStreamOutputStream>, tonic::Status> {
         let inner = request.into_inner();
         if let Some(console) = self.get_console(inner.console) {
-            Ok(tonic::Response::new(console.output_stream().await))
+            let stream = console
+                .output_stream()
+                .await
+                .map_err(|e| tonic::Status::internal(e.to_string()))?;
```
sjoerdsimons, line 753: still always returns an internal error I think? Better to implement `From<ConsoleError> for tonic::Status`  similar to e.g. `VolumeError,` then you can just return the error directly.

sjoerdsimons (changes requested):

boardswarm/src/main.rs
laeyraud: I forgot about this one (I focused on replacing the ConsoleError).
I'll apply your suggestion to convert ConsoleError to tonic::Status.

sjoerdsimons (approved):

Issue #87, opened 2024-05-06
cli: Track terminal size to enable greater console history to be shown when possible


mwelchuk:
At the moment boardswarm-cli assumes the a default 80 column by 24 row terminal size. It is generally possible to resize a terminal window when running from a GUI and I tend to do this for at least the height, so that I can see a greater depth of the console history.

I note that this isn't always well supported, I tend to use picocom and launching cut down versions of vi in it can result in the terminal sizing being lost (until vi is closed and the terminal is resized, which seems to update whatever information in the TTY that vi changed), however this can be really really useful when trying to follow verbose processes.

andrzejtp:
Actually, width is as important to me. Long lines which wrap look horrible and are difficult to read.

PR #153, opened 2024-12-31
Increase UI terminal size

laeyraud:
This merge requests aims to improve the terminal ergonomics in UI mode (see #87).

Previously, the console terminal size  was hardcoded to 80 characters x 24 lines. Now the UI size is auto detected and resized on the fly to fit the main terminal. 
A new option "--scrollback-lines" is also added to manually configure the scrollback buffer.

### DIFF
--- a/boardswarm-cli/src/main.rs
+++ b/boardswarm-cli/src/main.rs
@@ -712,6 +712,9 @@ enum Command {
         device: DeviceArg,
         #[command(flatten)]
         console: DeviceConsoleArgs,
+        #[clap(long, default_value_t = 5000)]
+        /// Number of lines to keep for scrollback
+        scrollback_lines: usize,
     },
 }
 
@@ -1335,12 +1338,16 @@ async fn main() -> anyhow::Result<()> {
             }
             Ok(())
         }
-        Command::Ui { device, console } => {
+        Command::Ui {
+            device,
+            console,
+            scrollback_lines,
+        } => {
             let device = device
                 .device(boardswarm)
                 .await?
                 .ok_or_else(|| anyhow::anyhow!("Device not found"))?;
-            ui::run_ui(device, console.console).await
+            ui::run_ui(device, console.console, scrollback_lines).await
         }
     }
 }
--- a/boardswarm-cli/src/main.rs
+++ b/boardswarm-cli/src/main.rs
@@ -33,6 +33,7 @@ use tokio::{
 use boardswarm_client::client::ItemEvent;
 use tokio_util::compat::{TokioAsyncReadCompatExt, TokioAsyncWriteCompatExt};
 use tracing::{debug, info};
+use ui::TerminalSizeSetting;
 use utils::BatchWriter;
 
 mod ui;
@@ -712,6 +713,11 @@ enum Command {
         device: DeviceArg,
         #[command(flatten)]
         console: DeviceConsoleArgs,
+        /// Terminal size, with the possible formats:
+        ///- "<columns>x<lines>" (ex: "72x18") where the terminal size is fixed
+        ///- "auto", where it is autoresized to match the window size
+        #[clap(long, default_value = "80x24", verbatim_doc_comment)]
+        terminal_size: TerminalSizeSetting,
         #[clap(long, default_value_t = 5000)]
         /// Number of lines to keep for scrollback
         scrollback_lines: usize,
@@ -1341,13 +1347,15 @@ async fn main() -> anyhow::Result<()> {
         Command::Ui {
             device,
             console,
+            terminal_size,
             scrollback_lines,
         } => {
             let device = device
                 .device(boardswarm)
                 .await?
                 .ok_or_else(|| anyhow::anyhow!("Device not found"))?;
-            ui::run_ui(device, console.console, scrollback_lines).await
+
+            ui::run_ui(device, console.console, terminal_size, scrollback_lines).await
         }
     }
 }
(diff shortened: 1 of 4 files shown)


sjoerdsimons (changes requested):
fwiw #87 would only be fixed by the first commit; the latter is seperate and could be merged seperately quite easily. 

The autoresizing is more controversial as *not* using the *local* terminal size is a feature not a bug. The fundamental problem is that what we're rendering is generated by the remote with an assumed terminal size; 80x24 is the default size for both u-boot as well as the typical default for linux console hence the current suggestion.  If remote and local understanding don't match you can get weird rendering artifacs. With u-boot this even happens on the console with long lines, once in linux it mostly show up for things that aren't simply line based output.

When you're using e.g. a local shell or e.g. ssh the columns/width is kept in sync (you can override them with e.g stty). For serial it's not that obvious as we can't be guaranteed the remote side recognize escape sequences..

fwiw xterm has an extension escape to resize the *local* size with the vt100 parse we use does recognize, so resizing can be triggered from both sides.

Overall I agree with the goal of removing the hardcoded 80x24, but it needs a bit more then just doing autoresizing.. As an mvp start we could add options to configure a static size and or optionally enable auto-resizing with documented caveats

laeyraud:
> fwiw #87 would only be fixed by the first commit; the latter is seperate and could be merged seperately quite easily.
> 
> The autoresizing is more controversial as _not_ using the _local_ terminal size is a feature not a bug. The fundamental problem is that what we're rendering is generated by the remote with an assumed terminal size; 80x24 is the default size for both u-boot as well as the typical default for linux console hence the current suggestion. If remote and local understanding don't match you can get weird rendering artifacs. With u-boot this even happens on the console with long lines, once in linux it mostly show up for things that aren't simply line based output.
> 
> When you're using e.g. a local shell or e.g. ssh the columns/width is kept in sync (you can override them with e.g stty). For serial it's not that obvious as we can't be guaranteed the remote side recognize escape sequences..
> 
> fwiw xterm has an extension escape to resize the _local_ size with the vt100 parse we use does recognize, so resizing can be triggered from both sides.
> 
> Overall I agree with the goal of removing the hardcoded 80x24, but it needs a bit more then just doing autoresizing.. As an mvp start we could add options to configure a static size and or optionally enable auto-resizing with documented caveats

I've reworked the commits to add as you suggested an option to select either a fixed value or being in autoresize. 
The default behaviour would remain the terminal size being in 80x24.

boardswarm-cli/src/main.rs
```diff
@@ -706,6 +706,12 @@ enum Command {
         device: DeviceArg,
         #[command(flatten)]
         console: DeviceConsoleArgs,
+        /// Terminal size, format: {columns}x{lines} (ex: 72x18)
+        #[clap(long, default_value = "80x24")]
```
sjoerdsimons, line 710: Any reason for tis not directly being `TerminalSizeSetting`? Also doesn't document the `auto` setting

sjoerdsimons (changes requested):

boardswarm-cli/src/main.rs
laeyraud: No reason, just forgot I could use the new type directly from clap parser. 
I modified the code to use it and also reword on the help to document auto option as well

boardswarm-cli/src/ui.rs
```diff
@@ -16,17 +18,76 @@ use crate::ui_term;
...
+impl FromStr for TerminalSizeSetting {
+    type Err = TerminalSizeSettingError;
+
+    fn from_str(fmt: &str) -> Result<Self, Self::Err> {
+        if fmt == "auto" {
+            return Ok(TerminalSizeSetting::Auto);
+        }
+
+        // Try to parse terminal size format to get width and height
+        if let Some(size_fmt) = fmt.split_once('x') {
```
sjoerdsimons, line 47: this is a nitpick but something like:
```rust
let (w, h) = fmt.split_once('x').ok_or(TerminalSizeSettingError::ParseError)?;
```
may be more elegant als it avoids the big if block where one case just returns an error; also unpacking tuples early is a bit more readable

boardswarm-cli/src/ui.rs
```diff
@@ -16,17 +18,76 @@ use crate::ui_term;
...
+            // Try to convert the first and second tuple values into u16
+            match (size_fmt.0.parse::<u16>(), size_fmt.1.parse::<u16>()) {
```
sjoerdsimons, line 49: If you implement `From ParseIntError for TerminalSizeSettingError` you can do quick return with `?` instead of having to do a match block here.

boardswarm-cli/src/ui.rs
```diff
@@ -16,17 +18,76 @@ use crate::ui_term;
...
         scrollback_lines: usize,
         tui: TuiTerminal<CrosstermBackend<std::io::Stdout>>,
     ) -> Self {
-        let parser = vt100::Parser::new(height, width, scrollback_lines);
-        let mut this = Self { parser, tui };
+        let size = match size_setting {
+            TerminalSizeSetting::Fixed(s) => s,
+            TerminalSizeSetting::Auto => match tui.size() {
+                Ok(s) => s,
+                Err(_) => Size {
```
sjoerdsimons, line 78: potentially add a warn! here about doing a fallback

boardswarm-cli/src/ui.rs
laeyraud: Fixed

boardswarm-cli/src/ui.rs
laeyraud: With this suggestion and the previous, it does simply a lot .

boardswarm-cli/src/ui.rs
laeyraud: Fixed

sjoerdsimons (approved):

PR #154, opened 2025-01-06
Add mouse support

### LINKED ISSUE #88: cli: Provide some kind of scroll mode in ui
Currently `boardswarm-cli ui` enables the console to be scrolled forward/backward one line at a time with `ctrl-a`, `j`|`k`. This works, however is really inefficient when trying to scroll back many lines.

How about providing a scrollback mode, maybe entered as soon as scrolling has been commenced, whereby the up/down arrow keys and maybe even page up/down can be used to move a line or some greater block at a time without needing to constantly type `ctrl-a`?

laeyraud:
This merge request adds the mouse support in the UI mode to fix #88

In order to support it, the UI mode now explicitly relies  on Crossterm crate as terminal backend (ratatui already uses it as default).  Instead of managing the stdin input manually, the keyboard and mouse actions are now listened through Crossterm event handler, filtered, processed and then sent (if needed) to the device input or output consoles.  

Note: the MR relies on the commits from GH-151 (the first 3 ones) so to be merged after it.

### DIFF
--- a/boardswarm-cli/src/ui.rs
+++ b/boardswarm-cli/src/ui.rs
@@ -1,5 +1,6 @@
 use std::task::Poll;
 
+use anyhow::anyhow;
 use bytes::Bytes;
 use futures::{pin_mut, ready, Stream, StreamExt};
 use ratatui::{
@@ -162,7 +163,11 @@ async fn run_ui_internal(
 
     let mut terminal = Terminal::new(80, 24, tui_terminal).await;
 
-    let output = console.clone().stream_output().await?;
+    let output = console
+        .clone()
+        .stream_output()
+        .await
+        .map_err(|e| anyhow!("Console stream opening failed: '{}'", e.message()))?;
 
     let (input_tx, input_rx) = tokio::sync::mpsc::channel(16);
     let _writer = tokio::spawn(async move {
--- a/boardswarm-cli/src/ui.rs
+++ b/boardswarm-cli/src/ui.rs
@@ -6,10 +6,12 @@ use ratatui::{
     backend::CrosstermBackend,
     layout::Rect,
     widgets::{Block, Borders},
-    Terminal as TuiTerminal,
+    DefaultTerminal, Terminal as TuiTerminal,
 };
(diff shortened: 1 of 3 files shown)


boardswarm-cli/src/ui.rs
```diff
@@ -189,68 +150,117 @@ async fn run_ui_internal(
...
-                        }
-                        Input::Up | Input::Down | Input::ScrollReset => {
-                            let _ = input_tx.send(i).await;
-                            None
-                        }
-                        Input::Bytes(data) => Some(data),
-                        Input::UIMessage(_) => None,
+                            Event::Mouse(mouse_event) => match mouse_event.kind {
+                                MouseEventKind::ScrollDown => {
+                                    let _ = input_tx.send(Input::Down).await;
```
sjoerdsimons, line 245: what if you're the remote console app responds to mouse events? e.g. when running htop remotely?

boardswarm-cli/src/ui.rs
```diff
@@ -70,79 +74,40 @@ impl Terminal {
...
+        KeyCode::PageDown => b"\x1B[6~".to_vec(),
+        KeyCode::Home => b"\x1B[1~".to_vec(),
+        KeyCode::End => b"\x1B[4~".to_vec(),
+        KeyCode::Up => b"\x1B[A".to_vec(),    // CUU
+        KeyCode::Down => b"\x1B[B".to_vec(),  // CUD
+        KeyCode::Left => b"\x1B[D".to_vec(),  // CUB
+        KeyCode::Right => b"\x1B[C".to_vec(), // CUF
+        KeyCode::Char('c') if modifiers == KeyModifiers::CONTROL => vec![0x3_u8], // ^C ETX
+        KeyCode::Char('d') if modifiers == KeyModifiers::CONTROL => vec![0x4_u8], // ^D EOT
+        KeyCode::Char(c) => {
```
sjoerdsimons, line 103: Rather then writing he ansi codes by hand; why not use crossterms Command where possible ? Or hopefully there are other crates to do the translation.

Also are you sure this is complete set?

Issue #160, opened 2025-02-05
Possible bug in boardswarm/src/mediatek_brom.rs ?


obbardc:
Hi!

The registration code for the mediatek_brom looks like this:

```
        if device.property_u64("ID_VENDOR_ID", 16) != Some(0x0e8d) {
            return false;
        };
        if device.property_u64("ID_MODEL", 16) != Some(0x0003) {
            return false;
        };
```

while writing a driver for another device, I notice that `ID_MODEL` should be `ID_MODEL_ID`, is it a bug in the mediatek_brom code ? Or does it happen to be that `ID_MODEL` is also `ID_MODEL_ID` for this device ?

In the latter case, I think it makes sense to change to `ID_MODEL_ID` to be more consistent. I am happy to write the patch if that's the case (but do not have hardware to confirm on)

sjoerdsimons:
Good catch; Just checking out what's going on here in more detail; Both properties are set by udev's usb builtin;  the _ID variants directly come from `idVendor` and `idProduct` so the actual VID and PID.  While the bare `ID_VENDOR` and `ID_MODEL`  variants first try to use `manufacturer` and `product` instead aka the optional Manufacturer and Product strings provided by the USB device, with a fallback to the VID/PID. 

So really we should be using the `_ID` variants, for well, ids. However as the mediatek bootrom on the devices I've tested with doesn't set the optional strings, it doesn't make a practical difference.

obbardc:
Yeah, that was exactly what I was thinking - it's just a cleanup thing really for existing devices.

It discovered this while creating a new driver for one of our custom devices - the VID/PID is set to the common VID/PID of the common development kit while the `manufacturer` and `product` string are set to the specific device. When I copied existing VID/PID detection code from another device, it didn't work straight away causing some head scratching ;-).

I created https://github.com/boardswarm/boardswarm/pull/161 based on your comment to rework it.

PR #161, opened 2025-02-06
Use ID_MODEL_ID instead of ID_MODEL for USB devices

obbardc:
Some USB products may set ID_MODEL to its own custom string but if its not set then udev sets ID_MODEL to ID_MODEL_ID (e.g. the idProduct, the actual device's PID).

Both properties are set by udev's usb builtin - except in the cases where ID_MODEL is set by the USB device to a custom string. The _ID variants directly come from idVendor and idProduct (e.g. the actual VID and PID), while the bare ID_VENDOR and ID_MODEL variants first try to use manufacturer and product instead (e.g. the optional Manufacturer and Product strings provided by the USB device) when a fallback to the VID/PID.

Rework the code to use the _ID variants where device IDs are used. This causes no functional change.

Fixes: https://github.com/boardswarm/boardswarm/issues/160

### DIFF
--- a/boardswarm/src/dfu.rs
+++ b/boardswarm/src/dfu.rs
@@ -49,6 +49,7 @@ pub async fn start_provider(name: String, server: Server) {
                     continue;
                 };
 
+                // Match on ID_MODEL here; the device may advertise the optional idModel property.
                 let name = if let Some(model) = device.property("ID_MODEL") {
                     format!("{}/{} {}", busnum, devnum, model)
                 } else {
--- a/boardswarm/src/mediatek_brom.rs
+++ b/boardswarm/src/mediatek_brom.rs
@@ -39,7 +39,7 @@ impl SerialProvider for MediatekBromProvider {
         if device.property_u64("ID_VENDOR_ID", 16) != Some(0x0e8d) {
             return false;
         };
-        if device.property_u64("ID_MODEL", 16) != Some(0x0003) {
+        if device.property_u64("ID_MODEL_ID", 16) != Some(0x0003) {
             return false;
         };
 
--- a/boardswarm/src/rockusb.rs
+++ b/boardswarm/src/rockusb.rs
@@ -47,7 +47,7 @@ pub async fn start_provider(name: String, server: Server) {
                     continue;
                 };
 
-                let name = if let Some(model) = device.property("ID_MODEL") {
+                let name = if let Some(model) = device.property("ID_MODEL_ID") {
                     format!("{}/{} {}", busnum, devnum, model)
                 } else {
                     format!("{}/{}", devnum, devnum)


boardswarm/src/dfu.rs
```diff
@@ -49,7 +49,7 @@ pub async fn start_provider(name: String, server: Server) {
                     continue;
                 };
 
-                let name = if let Some(model) = device.property("ID_MODEL") {
+                let name = if let Some(model) = device.property("ID_MODEL_ID") {
```
obbardc, line 52: Untested but should work.

boardswarm/src/mediatek_brom.rs
```diff
@@ -39,7 +39,7 @@ impl SerialProvider for MediatekBromProvider {
         if device.property_u64("ID_VENDOR_ID", 16) != Some(0x0e8d) {
             return false;
         };
-        if device.property_u64("ID_MODEL", 16) != Some(0x0003) {
+        if device.property_u64("ID_MODEL_ID", 16) != Some(0x0003) {
```
obbardc, line 42: Untested but should work.

boardswarm/src/rockusb.rs
```diff
@@ -47,7 +47,7 @@ pub async fn start_provider(name: String, server: Server) {
                     continue;
                 };
 
-                let name = if let Some(model) = device.property("ID_MODEL") {
+                let name = if let Some(model) = device.property("ID_MODEL_ID") {
```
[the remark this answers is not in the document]
obbardc: OK, this is still the devices pid (at least for RK3588 brom):
```
INFO start_provider{name="rockusb"}: boardswarm::rockusb: New rockusb volume: 5/17 350b
```

boardswarm/src/dfu.rs
sjoerdsimons: iirc for those ID_MODEL might actaully be correct as it's used for the name rather then being assumped to be an id

boardswarm/src/rockusb.rs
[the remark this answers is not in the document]
sjoerdsimons: yeah that's another one that can go either way as it's used for naming; though maskrom is unlikely to have a string indeed

boardswarm/src/mediatek_brom.rs
sjoerdsimons: looks correct; something @laeyraud can easily test

boardswarm/src/dfu.rs
obbardc: ok great, i can quickly revert that hunk but i think it makes sense to be sure first by checking a device ?

boardswarm/src/mediatek_brom.rs
obbardc: if you could test it would be worthwhile, since i dont want to introduce a regression :-)

laeyraud: I tested with both MTK Genio 510 and 1200 boards and it seems OK.

This modification also makes sense to me as @sjoerdsimons  recommended  me and others  `ID_MODEL_ID` in matching rules for other providers in configuration.

boardswarm/src/dfu.rs
obbardc: Updated.

boardswarm/src/dfu.rs
```diff
@@ -49,6 +49,7 @@ pub async fn start_provider(name: String, server: Server) {
                     continue;
                 };
 
+                // Match on ID_MODEL here; the device may adversite the optional idModel property.
```
sjoerdsimons, line 52:
```suggestion
                // Match on ID_MODEL here; the device may advertise the optional idModel property.
```

PR #163, opened 2025-02-10
Add support for Linaro QCOMLT Debug Board

obbardc:
The Linaro Debug Board[0], also known as the QCOMLT Debug Board (where QCOMLT expands to Qualcomm Landing Team) is a small, open-source debug board which was designed to plug directly into various Qualcomm development kits and provide functionality allowing developers to debug the DUTs in a generic/remote-first way (e.g. provide ability to switch DUT power, measure DUT power, switch USB connection on/off to prevent back-feeding power and to put the DUT into various states by simulating button presses).

The debug board is currently used with cdba[1] as a way to provide remote access to many boards to developers. cdba is also used as a layer under GitLab CI to allow testing of Linux kernel and images under CI for some projects.

Unfortunately cbda doesn't scale too well and uses ssh as its communication backend (which amongst other problems lacks a granulated external authentication) so we would like to, over the long term, switch to a new tool for remote access. Boardswarm is a good candidate.

---

The Linaro QCOMLT Debug Board has an Atmel MCU with firmware which presents itself as two USB-CDC serial ports. Each device has a unique SerialNumber but uses generic Atmel product and vendor IDs.

The board enumerates in Linux as:

    usb 1-2: new full-speed USB device number 106 using xhci_hcd
    usb 1-2: New USB device found, idVendor=03eb, idProduct=2404, bcdDevice= 1.00
    usb 1-2: New USB device strings: Mfr=1, Product=2, SerialNumber=3
    usb 1-2: Product: DebugBoard
    usb 1-2: Manufacturer: Linaro
    usb 1-2: SerialNumber: 75eaaac075a445150202022343a460ff
    cdc_acm 1-2:1.0: ttyACM0: USB ACM device
    cdc_acm 1-2:1.2: ttyACM1: USB ACM device

The first enumerated serial port is used for controlling the Debug Board (e.g. controlling DUT power), while the second serial port is connected to the DUT as UART.

The support for the DUT UART port is handled by the existing serial provider, while support for the various functions exposed by the Debug Board's internal UART are exposed in boardswarm as Actuators.

There is also a separate function which allows getting the voltage/current drawn by the DUT; this hasn't yet been hooked up as there is no way (yet) in boardswarm to register measurement device(s), but have left TODOs in the code around this. Also, see [3].

[0]: Linaro Debug Board: https://git.codelinaro.org/linaro/qcomlt/debugboard
[1]: cdba: http://github.com/linux-msm/cdba
[3]: https://github.com/boardswarm/boardswarm/issues/164

Please note this is VERY early prototype, which does work, created for additional feedback.

### DIFF
--- a/boardswarm/src/main.rs
+++ b/boardswarm/src/main.rs
@@ -13,6 +13,7 @@ use futures::prelude::*;
 use futures::stream::BoxStream;
 use hifive_p550_mcu::HifiveP550MCUProvider;
 use mediatek_brom::MediatekBromProvider;
+use qcomlt_debug_board::QCOMLTDebugBoardProvider;
 use registry::{Properties, Registry, RegistryIndex};
 use std::fmt::Display;
 use std::net::{AddrParseError, SocketAddr};
@@ -38,6 +39,7 @@ mod gpio;
 mod hifive_p550_mcu;
 mod mediatek_brom;
 mod pdudaemon;
+mod qcomlt_debug_board;
 mod registry;
 mod rockusb;
 mod serial;
@@ -1163,6 +1165,14 @@ async fn main() -> anyhow::Result<()> {
                 )),
                 None => {
                     bail!("Mediatek brom provider requires the serial provider to be enabled")
+                }
+            },
+            qcomlt_debug_board::PROVIDER => match serial {
+                Some(ref s) => {
+                    s.add_provider(QCOMLTDebugBoardProvider::new(p.name, server.clone()))
+                }
+                None => {
+                    bail!("QCOMLT Debug Board provider requires the serial provider to be enabled")
                 }
             },
             rockusb::PROVIDER => {
--- a/boardswarm/src/qcomlt_debug_board.rs
+++ b/boardswarm/src/qcomlt_debug_board.rs
@@ -1,0 +1,244 @@
+use std::{collections::HashMap, path::PathBuf};
+
+use serde::Deserialize;
+use tokio::io::AsyncWriteExt;
+use tokio::sync::oneshot;
+use tokio_serial::SerialPortBuilderExt;
+use tracing::{info, warn};
+
+use crate::{ActuatorError, ActuatorId, Server, registry, serial::SerialProvider, udev::Device};
+
+pub const PROVIDER: &str = "qcomlt-debug-board";
+
+pub struct QCOMLTDebugBoardProvider {
+    name: String,
+    server: Server,
+    registrations: HashMap<PathBuf, Vec<ActuatorId>>,
+}
+
+impl QCOMLTDebugBoardProvider {
+    pub fn new(name: String, server: Server) -> Self {
+        Self {
+            name,
+            server,
+            registrations: Default::default(),
+        }
+    }
+}
+
+impl SerialProvider for QCOMLTDebugBoardProvider {
+    fn handle(&mut self, device: &Device, _seqnum: u64) -> bool {
+        let provider_properties = &[
+            (registry::PROVIDER_NAME, self.name.as_str()),
+            (registry::PROVIDER, PROVIDER),
+        ];
+
+        /* QCOMLT Debug Board uses the default Atmel devkit VID/PIDs
+        but advertises a custom ID_VENDOR and ID_MODEL. */
+        if device.property_u64("ID_VENDOR_ID", 16) != Some(0x03eb) {
+            return false;
+        };
+        if device.property_u64("ID_MODEL_ID", 16) != Some(0x2404) {
+            return false;
+        };
+        if device.property("ID_VENDOR") != Some("Linaro") {
+            return false;
+        };
+        if device.property("ID_MODEL") != Some("DebugBoard") {
+            return false;
+        };
+
+        // We only care about the debug board's "management" serial port, e.g.
+        // the first serial port the device enumerates. This is used for functions
+        // such as turning the DUT power on, enabling USB VBUS, pressing buttons.
+        //
+        // The second serial port is used for the DUT console - that
+        // can be used by the regular serial provider (and matching by
+        // ID_USB_SERIAL_SHORT)
+        if device.property("ID_USB_INTERFACE_NUM") != Some("00") {
+            return false;
+        };
+
+        if let Some(node) = device.devnode()
+            && let Some(name) = node.file_name()
+        {
+            let Some(serial_number) = device.property("ID_USB_SERIAL_SHORT") else {
+                warn!(
+                    "Skipping QCOMLT Debug Board on {}: missing ID_USB_SERIAL_SHORT",
+                    node.display()
+                );
+                return false;
+            };
+
+            info!(
+                "Reserving console {} for QCOMLT Debug Board {}",
+                node.display(),
+                serial_number
+            );
+
+            let port = match tokio_serial::new(node.to_string_lossy(), 115200).open_native_async() {
+                Ok(port) => port,
+                Err(e) => {
+                    warn!("Failed to open QCOMLT Debug Board management serial port: {e}");
+                    return false;
+                }
+            };
+
+            let mut base_properties = device.properties(name.to_string_lossy());
+            base_properties.extend(provider_properties);
+            base_properties.insert("qcomlt_debug_board.serial", serial_number);
+
+            let (tx, exec) = tokio::sync::mpsc::channel(16);
+            tokio::spawn(process(port, exec));
+
+            let actuator_types = [
+                ("power", QCOMLTActuatorType::Power),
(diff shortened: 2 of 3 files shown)


boardswarm/src/qcomlt_debug_board.rs
```diff
@@ -0,0 +1,321 @@
...
+    mut exec: tokio::sync::mpsc::Receiver<QCOMLTDebugBoardCommand>,
+) {
+    while let Some(command) = exec.recv().await {
+        match command {
+            /* TODO: Create a standalone rust library for the QCOMLTDebugBoard
+            similar to https://github.com/boardswarm/mediatek-brom rather than
+            sending raw bytes here. */
+            QCOMLTDebugBoardCommand::PowerOn(sender) => {
+                info!("Sending PowerOn command");
+                let r = port.write_all(b"P\n").await;
```
obbardc, line 257: Not sure if newline needed. Need to check.

boardswarm/src/qcomlt_debug_board.rs
```diff
@@ -0,0 +1,321 @@
...
+        let (device, exec) = tokio::sync::mpsc::channel(16);
+        tokio::spawn(process(port, exec));
+        Self { device, name }
+    }
+}
+
+/*
+TODO: calling change-mode in cli directly does not work (but setting mode does). Does it also not work for the other implementations ?
+
+obbardc@nox:~/projects/git/boardswarm$ RUST_LOG=info cargo run --bin boardswarm-cli -- -i localhost actuator 1 change-mode on
```
sjoerdsimons, line 286: actuators don't have a mode change; they take a json string to do their settings, similar to what you configure in the device section wrt. modes. This is to support differnet types of actuators especially ones that aren't just boolean :)

sjoerdsimons:
This looks pretty cool; I've converted it to draft state for now because, as you say yourself it's a very early prototype. But i'm quite happy to help you support this, especially given it's open hardware. 

Wrt. to your notes, could you open an issue for allowing bulk registrations, so we can track that seperately? 

I didn't do a detailed review of your code just yet, it's obviously too early :) But generally it seems to be going the right direction. 

Fwiw one pattern I did for most other things is to have separate crates for the control protocols. As that makes it both easier to create other tools and to test/maintain  the implementation separately from boardswarm. That might be something to consider. That said looking at the firmware source, it seems it only recognizes 9 commands at the moment. So i'd be happy either way.

boardswarm/src/qcomlt_debug_board.rs
sjoerdsimons: judging by the firmware code you don't.

boardswarm/src/qcomlt_debug_board.rs
obbardc: OH, change-mode expects json format:

```
Change actuator mode

Usage: boardswarm-cli actuator <ACTUATOR> change-mode <MODE>

Arguments:
  <MODE>  Actuator specific mode in json format

Options:
  -h, --help  Print help
```

sjoerdsimons: that's what i said! `change-mode` might not be the best phrasing, but typically that's purely internal.. Conceptually the actuator only actuates as requested, it's only a device that knows what it needs for a given mode.

Fwiw atm it's a bit silly as we just have GPIO's; However for a controllable bench supply rather then just on/off you'd configure a few more parameters (e.g. voltage, over current protection etc).

obbardc: Right, but passing JSON in isn't clear and the response when you pass e.g. `on` instead of json string doesn't really match. So I will open an issue/PR to clarify that :-)

obbardc: fwiw for the record this was reported as !165

obbardc:
> This looks pretty cool; I've converted it to draft state for now because, as you say yourself it's a very early prototype. But i'm quite happy to help you support this, especially given it's open hardware.

Thanks. We have limited boards, but I may be able to get one for you to test at some point. Although it's pretty useless if you don't have interest in QC devkits :-)


> Wrt. to your notes, could you open an issue for allowing bulk registrations, so we can track that seperately?

I created https://github.com/boardswarm/boardswarm/issues/167, hope that makes sense.


> I didn't do a detailed review of your code just yet, it's obviously too early :) But generally it seems to be going the right direction.

Yeah, as you know my rust is pretty awful and this is obviously just lashed up, but my goal is get something working initially and simply opening a PR keeps the code in the open too :-)


> Fwiw one pattern I did for most other things is to have separate crates for the control protocols. As that makes it both easier to create other tools and to test/maintain the implementation separately from boardswarm. That might be something to consider. That said looking at the firmware source, it seems it only recognizes 9 commands at the moment. So i'd be happy either way.

Yeah, I was thinking for now to simply include it in the one file for ease; I created a TODO in the comment to split it out into its own crate. I would like to do that once support is further, it makes sense to do that when I add support for e.g voltage/current measurements.

obbardc:
@sjoerdsimons The blocker for me here is to be able to register multiple actuators for this device and the `ActuatorRegistrationState` code which I blindly copied. If you can help direct me with these bits at some point it'd be appreciated.

boardswarm/src/qcomlt_debug_board.rs
obbardc: fixed that.

boardswarm/src/main.rs
```diff
@@ -32,6 +33,7 @@ mod fastboot;
 mod gpio;
 mod mediatek_brom;
 mod pdudaemon;
+mod qcomlt_debug_board;
```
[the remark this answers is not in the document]
obbardc: @sjoerdsimons any preference for the name ? I really think `qcomlt debug board` for the hardware was a bit shortsighted; but meh we are stuck with it now.

boardswarm/src/qcomlt_debug_board.rs
```diff
@@ -0,0 +1,321 @@
...
+    PowerOff(tokio::sync::oneshot::Sender<Result<(), std::io::Error>>),
+}
+
+async fn process(
+    mut port: tokio_serial::SerialStream,
+    mut exec: tokio::sync::mpsc::Receiver<QCOMLTDebugBoardCommand>,
+) {
+    while let Some(command) = exec.recv().await {
+        match command {
+            /* TODO: Create a standalone rust library for the QCOMLTDebugBoard
```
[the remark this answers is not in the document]
obbardc: For now, I will leave this code in here until we are happy with the boardswarm implementation, then split out to separate crate.

boardswarm/src/qcomlt_debug_board.rs
```diff
@@ -0,0 +1,318 @@
...
+        false
+    }
+
+    fn remove(&mut self, device: &Device) {
+        self.registrations.remove(device.syspath());
+    }
+}
+
+#[instrument(skip(r, properties))]
+async fn setup_subdevices(
```
sjoerdsimons, line 190: Looking at this code again in the context of improving the DeviceRegistrations helper; In your case you really don't need this function, you can simply register within QCOMLTDebugBoardProvider::handle directly.

The background for the Devicereigstration helper was because for mediatek bootrom i'm doing some extra probing after opening the device to e.g. get serial information. Which i wanted to do in another async task to not block register new serial based device while talking to the mtk bootrom. However that then opens a race window where the serial port could bounce while we're trying to talk to it, which potentially could lead to a device getting registered twice (tbh. there is still a tiny window as the second probe might fail due to the first probe accidentally opening the node).

For your code you can can simply use a hashmap to keep a device to id (or id's) mapping and directly call the relevant `register_*` functions  in `SerialProvider::handle` and `unregister_*`  in `SerialProvider::remove`

obbardc: I've fixed this now after rebasing on main.

obbardc:
@sjoerdsimons can I ask for another review on this now I've rebased onto main and fixed the device registering bits ? I will look in the next week or so at fixing the remaining TODOs already in the code.

boardswarm/src/qcomlt_debug_board.rs
```diff
@@ -0,0 +1,201 @@
...
+                "Reserving console {} for QCOMLT Debug Board {}",
+                node.display(),
+                serial_number
+            );
+
+            let port = match tokio_serial::new(node.to_string_lossy(), 115200).open_native_async() {
+                Ok(port) => port,
+                Err(e) => {
+                    warn!("Failed to open serial port: {e}");
+                    return false;
```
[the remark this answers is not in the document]
obbardc: Not sure about this. Think returning false is the right thing to do here.

PR #169, opened 2025-02-19
Build Docker containers for arm64 architecture

obbardc:
In some cases wee are using arm64 devices to control DUTs. Build the containers in CI for arm64 platforms (as well as existing amd64 containers).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,9 +88,15 @@ jobs:
     permissions:
       contents: read
       packages: write
+    env:
+      DOCKER_TARGET_PLATFORMS: linux/amd64,linux/arm64
     steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
       - name: Setup Docker buildx
         uses: docker/setup-buildx-action@v3
+        with:
+          platforms: ${{ env.DOCKER_TARGET_PLATFORMS }}
       - uses: actions/checkout@v4
       - name: Extract Docker metadata
         id: meta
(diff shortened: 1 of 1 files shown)


obbardc:
(untested for now; awaiting CI results)

obbardc:
OK, this causes CI (for boardswarm docker container ) to go from [2:58](https://github.com/boardswarm/boardswarm/actions/runs/13346087262/job/37276713645) to [over 30 minutes](https://github.com/boardswarm/boardswarm/actions/runs/13457156275/job/37603714740?pr=169), of course due to building in qemu cross-arch. I don't think it's worth it - I will switch to using the prebuilt binaries on arm64 instead of a container...

sjoerdsimons:
> OK, this causes CI (for boardswarm docker container ) to go from [2:58](https://github.com/boardswarm/boardswarm/actions/runs/13346087262/job/37276713645) to [over 30 minutes](https://github.com/boardswarm/boardswarm/actions/runs/13457156275/job/37603714740?pr=169), of course due to building in qemu cross-arch. I don't think it's worth it - I will switch to using the prebuilt binaries on arm64 instead of a container...

Yeah little point in emulating the compiler :)

obbardc:
Unless you have any suggestions I'm happy to close this.

sjoerdsimons:
Fwiw I tend to only use the docker container for instances that purely are boardswarm clients; On the arm64 controllers I just use the prebuild binaries :) Fwiw there are arm64 runners available as well; but I don't think you  can easily build multi-arch containers in multiple runs.  I do wonder if you can have the builder container be a seperate arch from the final stage build

obbardc:
let's just close this for now then; i have no reason to work on it, as in previous comment, I will use the prebuilt boardswarm binaries now :-)

PR #175, opened 2025-03-28
Add details about providers not currently in documentation

mwelchuk:
The fastboot and mediatek-brom providers are not currently listed in the documentation, add sections for these providers.

### DIFF
--- a/boardswarm/README.md
+++ b/boardswarm/README.md
@@ -177,6 +177,20 @@ provider:
         - mmc0boot1
 ```
 
+### MediaTek boot ROM provider (mediatek-brom)
+
+Support for MediaTek boot ROM protocol. mediatek-brom devices are automatically
+detected by scanning for USB serial devices with vendor ID `0x0e8d` and
+product ID `0x0003`. No provider specific parameters are expected and only one
+of this provider can exist.
+
+Example configuration:
+```
+provider:
+  - name: mediatek-brom
+    provider: mediatek-brom
+```
+
 ### Rock USB provider (rockusb)
 
 Support for rockchip USB protocol. rockusb devices are autodetected
--- a/boardswarm/README.md
+++ b/boardswarm/README.md
@@ -152,6 +152,31 @@ provider:
     provider: dfu
 ```
 
+### Fastboot provider (fastboot)
+
+Support for the fastboot communication protocol exposed over USB. As fastboot
+partitions aren't necessarily discoverable this can be configured to match
+certain devices and set a preconfigured set of targets, though autodetection
+will be attempted.
+
+Example configuration:
+```
+provider:
+  - name: fastboot
+    provider: fastboot
+    parameters:
+      match:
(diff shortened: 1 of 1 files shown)


boardswarm/README.md
```diff
@@ -152,6 +152,46 @@ provider:
...
+        - mmc0
+        - mmc0boot0
+        - mmc0boot1
+```
+
+### MediaTek boot ROM provider (mediatek-brom)
+
+Support for MediaTek boot ROM protocol. mediatek-brom devices are automatically
+detected by scanning for USB serial devices with vendor id 0x0e8d and product id
+0x0003. No provider specific parameters are expected and only one of this
```
obbardc, line 183:
```suggestion
`0x0003`. No provider specific parameters are expected and only one of this
```

boardswarm/README.md
```diff
@@ -152,6 +152,46 @@ provider:
...
+provider can exist.
+
+Currently this provider only supports the bootrom protocol over serial.
```
obbardc, line 186: This is confusing to me, what does `protocol over serial` mean ?

mwelchuk: I was under the impression that the MediaTek bootrom protocol could be used over other communication mediums, primarily based on the comment in the example config file:

https://github.com/boardswarm/boardswarm/blob/f5e849313d071f1f69e2e247a4d3bef84696efe1/boardswarm/share/server.conf#L52

Though to be honest, I can't find any information to suggest what other options might be available, I wonder if @sjoerdsimons would be able to elaborate.

sjoerdsimons: boardswarm only supports it over serial ports (as exposed by the bootrom of genio devices over usb); As i understand  the same protocol can be used about other transports as well, but i don't have experience with those devices :)  

And for extra fun it seems on openwrt one the bootrom sends some extra serial data (judging by some of their uart loading implementation) as a banner.. So worth noting that this is only tested on genio devices for now (Still on the fence whether i should order an openwrt one)

boardswarm/README.md
```diff
@@ -152,6 +152,46 @@ provider:
...
+
+Example configuration:
+```
+provider:
+  - name: fastboot
+    provider: fastboot
+    parameters:
+      match:
+        udev.ID_VENDOR_ID: 0e8d
+        udev.ID_MODEL_ID: 201c
```
sjoerdsimons, line 170: worth noting this the vendor/model for at least genio devices

boardswarm/README.md
sjoerdsimons: That said; as per @obbardc confusion this line is redundant as you say it already scans for USB serial devices :)

mwelchuk: Removed.

boardswarm/README.md
mwelchuk: Added.

sjoerdsimons:
closing as this was merged via #203

PR #187, opened 2025-05-22
boardswarm: fastboot: Implement Android's fastboot stage command

adalessandro:


boardswarm-protocol/proto/boardswarm.proto
```diff
@@ -23,7 +23,7 @@ service Boardswarm {
   rpc VolumeIo (stream VolumeIoRequest) returns (stream VolumeIoReply);
 
   // Commit all data in an uploader specific manner; e.g. may trigger a USB reset for DFU devices
-  rpc VolumeCommit(VolumeRequest) returns (google.protobuf.Empty);
+  rpc VolumeCommit(VolumeCommitRequest) returns (google.protobuf.Empty);
```
sjoerdsimons, line 26: I really need to pin down the semantics for commit somewhat better; (and ditto shutdown wrt. the requests). As it's getting somewhat confusing.. But generally it was meant as finish-everything-the-volume is doing hence their on purpose not being a volume target.

For the `continue`  on fastboot this seems very much a case of committing in that sense as well as there is no specific target per se; It is meant to continue the normal boot flow. As you can also see from your next commit which ignores the target entirely when running continue :)

boardswarm/src/fastboot.rs
```diff
@@ -18,6 +18,12 @@ use crate::{
 
 pub const PROVIDER: &str = "fastboot";
 
+/// Fastboot volume target to implement Android's fastboot stage command
+///
+/// Writing to this volume target will perform a download, but skip the flashing
+/// step. A commit to this target will perform a continue command.
+pub const TARGET_STAGE: &str = "stage";
```
sjoerdsimons, line 25: Does it make sense to hardcode this or should it be a configuration parameter for the fastboot provider instead? (similar to the initial list of targets)

sjoerdsimons (changes requested):
Fwiw for things depending on e.g. unrelased fastboot, feel free to mark the PR as draft and have a cargo patch to use your git branch of fastboot. That way we can at least have CI run properly

boardswarm/src/fastboot.rs
adalessandro: @sjoerdsimons , just to confirm, what you suggest is adding a configuration parameter to set the name of the volume target for the stage command? Which, in this commit, is hardcoded to `stage`. If that's your suggestion, we could have the following fastboot configuration in `server.conf`:
```
  - name: genio-fastboot                                                        
    provider: fastboot                                                          
    parameters:                                                                 
      match:                                                                    
        udev.ID_VENDOR_ID: "0e8d"                                               
        udev.ID_MODEL_ID: "201c"                                                
      targets:                                                                  
        - mmc0                                                                  
        - mmc0boot0                                                             
        - mmc0boot1                                                             
      stage: "stage"                                                            
```
That'd be equivalent to the `stage` target name hardcoded.
That being said, I don't really see a big win on having this stage volume target name configurable,
as it is only to somehow match the target name, in order to perform a download/continue, instead of a download/flash sequence.

boardswarm-protocol/proto/boardswarm.proto
adalessandro: > I really need to pin down the semantics for commit somewhat better; (and ditto shutdown wrt. the requests). As it's getting somewhat confusing.. But generally it was meant as finish-everything-the-volume is doing hence their on purpose not being a volume target.
> 
> For the `continue` on fastboot this seems very much a case of committing in that sense as well as there is no specific target per se; It is meant to continue the normal boot flow. As you can also see from your next commit which ignores the target entirely when running continue :)

Fastboot in the device ignores the target, but the `commit` behavior is different if the target volume is the special `stage` or any other target.

boardswarm/src/fastboot.rs
adalessandro: Added a configurable boolean option `stage`, e.g.:
```
  - name: genio-fastboot                                                        
    provider: fastboot                                                          
    parameters:                                                                 
      match:                                                                    
        udev.ID_VENDOR_ID: "0e8d"                                               
        udev.ID_MODEL_ID: "201c"                                                
      targets:                                                                  
        - mmc0                                                                  
        - mmc0boot0                                                             
        - mmc0boot1                                                             
      stage: true                                                               
```

boardswarm/src/fastboot.rs
```diff
@@ -18,12 +18,19 @@ use crate::{
...
+/// perform a download, but skip the flashing step.
+pub const TARGET_STAGE: &str = "stage";
+
 #[derive(Deserialize, Debug, Default)]
 struct FastbootParameters {
     #[serde(rename = "match")]
     match_: HashMap<String, String>,
     #[serde(default)]
     targets: Vec<String>,
+    stage: bool,
```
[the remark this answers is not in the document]
adalessandro: You're right. Fixed.

sjoerdsimons (approved):

PR #196, opened 2025-06-22
Improve udev registration helper

sjoerdsimons:
the registration helper was rather limited; Only volume's could be registered with it and at most one of those. Extend that so it can be used for any type as well as supporting multiple items to be registered in a batch

This also refactors the registry indexing so they can be typed rather then just u64, which makes it very explicit what an id refers to.

### DIFF
--- a/boardswarm/src/config_device.rs
+++ b/boardswarm/src/config_device.rs
@@ -33,7 +33,7 @@ impl DeviceNotifier {
 struct DeviceMode {
     name: String,
     depends: Option<String>,
-    sequence: Vec<DeviceItem<crate::config::ModeStep>>,
+    sequence: Vec<DeviceItem<u64, crate::config::ModeStep>>,
 }
 
 impl From<crate::config::Mode> for DeviceMode {
@@ -47,14 +47,15 @@ impl From<crate::config::Mode> for DeviceMode {
     }
 }
 
-struct DeviceItem<C> {
+struct DeviceItem<I, C> {
+    id: Mutex<Option<I>>,
     config: C,
-    id: Mutex<Option<u64>>,
 }
 
-impl<C> DeviceItem<C>
+impl<C, I> DeviceItem<I, C>
 where
     C: DeviceConfigItem,
+    I: Copy + Eq,
 {
     fn new(config: C) -> Self {
         Self {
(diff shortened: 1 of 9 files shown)


boardswarm/src/config_device.rs
```diff
@@ -67,15 +68,18 @@ where
...
 
-    fn get(&self) -> Option<u64> {
+    fn get(&self) -> Option<I> {
         *self.id.lock().unwrap()
     }
 
-    fn unset_if_matches(&self, id: u64) -> bool {
+    fn unset_if_matches(&self, id: I) -> bool
+    where
+        I: Eq,
```
adalessandro: This could be dropped, trait bound already in DeviceItem where clause.

boardswarm/src/config_device.rs
```diff
@@ -86,7 +90,10 @@ where
         }
     }
 
-    fn set_if_matches(&self, id: u64, properties: &Properties) -> bool {
+    fn set_if_matches(&self, id: I, properties: &Properties) -> bool
+    where
+        I: Eq,
```
adalessandro, line 94: This could be dropped, trait bound already in DeviceItem where clause.

boardswarm/src/main.rs
```diff
@@ -423,13 +424,43 @@ trait Device: Send + Sync {
...
 }
 
+macro_rules! impl_u64_index {
+    ($id:ident) => {
+        #[derive(Copy, Clone, Debug, Default, Hash, PartialOrd, Ord, PartialEq, Eq)]
+        pub struct $id(u64);
+
+        impl Display for $id {
+            fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                write!(f, "Actuator {}", self.0)
```
adalessandro, line 434: The string here was left hardcoded to "Actuator"

adalessandro:
With the simple comments above, PR approved.

boardswarm/src/main.rs
sjoerdsimons: Nice catch, should be fixed now :)

adalessandro: Ack, I think it's ready to be merged then.

PR #202, opened 2025-06-26
boardswarm: Add hifive-p550-storage provider

adalessandro:
```
    boardswarm: Add eswin-eic7700-storage provider
    
    The eswin-eic7700-storage provider registers a volume with a single
    target `bootloader`.
    
    When in USB download mode, the ESWIN EIC7700X will register itself as a
    mass storage device. When the `bootloader_ddr5_secboot.bin` bootloader
    artifact is written to its filesystem, it will automatically unregister
    and bootstrap into this bootloader image.
    
    Writing to the volume target `bootloader` will create and write the
    content to this file `bootloader_ddr5_secboot.bin`.
    
    Example server configuration (HiFive Premier P550 Development Board):
    ```
    providers:
      - name: eswin-eic7700-storage
        provider: eswin-eic7700-storage
        parameters:
          match:
            udev.ID_VENDOR: "ESWIN"
            udev.ID_MODEL: "WIN-2030"
          mountpoint: "/var/run/boardswarm/eswin-eic7700-storage"
    
    devices:
      - name: sifive-hifive-p550-evk-cbg-0
        consoles:
        volumes:
          - name: storage
            match:
              udev.ID_PATH: "pci-0000:02:00.2-usb-0:1.1.2:1.0-scsi-0:0:0:0"
    ```
    
    The user running boardswarm need to have access to the directory where
    the storage device is mounted.
    
    For example, a udev rule like the following could be added to automount
    the device using systemd-mount:
    ```
    $ cat /etc/udev/rules.d/99-boardswarm.rules
    ACTION=="add", SUBSYSTEMS=="usb", SUBSYSTEM=="block", \
        ENV{ID_VENDOR}=="ESWIN", ENV{ID_MODEL}=="WIN-2030", TAG+="systemd", \
        RUN{program}+="/usr/bin/systemd-mount --owner=boardswarm --no-block --automount=no --collect \
        $devnode /var/run/boardswarm/eswin-eic7700-storage/$env{ID_PATH}"
    ```
    
    Signed-off-by: Ariel D'Alessandro <ariel.dalessandro@collabora.com>
```

### DIFF
--- a/boardswarm/Cargo.toml
+++ b/boardswarm/Cargo.toml
@@ -45,3 +45,4 @@ reqwest = { version = "0.12.15", default-features = false, features = ["rustls-t
 strum = { version = "0.27.1", features = ["derive"] }
 strum_macros = "0.27.1"
 regex = "1.11.1"
+proc-mounts = "0.3.0"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -343,6 +343,7 @@ dependencies = [
  "mediatek-brom",
  "nusb",
  "pdudaemon-client",
+ "proc-mounts",
  "regex",
  "reqwest",
  "rockusb",
@@ -888,7 +889,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "976dd42dc7e85965fe702eb8164f21f450704bdde31faefd6471dba214cb594e"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2155,6 +2156,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "partition-identity"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9fa925f9becb532d758b0014b472c576869910929cf4c3f8054b386f19ab9e21"
+dependencies = [
+ "thiserror 1.0.69",
+]
+
+[[package]]
 name = "paste"
 version = "1.0.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2399,6 +2409,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "proc-mounts"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0d652f8435d0ab70bf4f3590a6a851d59604831a458086541b95238cc51ffcf2"
+dependencies = [
+ "partition-identity",
+]
+
+[[package]]
 name = "prost"
 version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2535,7 +2554,7 @@ dependencies = [
  "once_cell",
  "socket2",
  "tracing",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2833,7 +2852,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -2846,7 +2865,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.9.4",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -3334,7 +3353,7 @@ dependencies = [
  "getrandom 0.3.2",
  "once_cell",
  "rustix 1.0.7",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
--- a/boardswarm/src/eswin_eic7700_storage.rs
+++ b/boardswarm/src/eswin_eic7700_storage.rs
@@ -1,0 +1,230 @@
+use bytes::Bytes;
+use proc_mounts::{MountInfo, MountIter};
+use serde::Deserialize;
+use std::{
+    collections::HashMap,
+    io::SeekFrom,
+    path::{Path, PathBuf},
+    time::Duration,
+};
+use tokio::io::{AsyncSeekExt, AsyncWriteExt};
+use tokio_stream::StreamExt;
+use tracing::{debug, instrument, trace, warn};
+
+use crate::{
+    registry::{self, Properties},
+    udev::{DeviceEvent, DeviceRegistrations, PreRegistration},
+    Server, Volume, VolumeError, VolumeTarget, VolumeTargetInfo,
+};
+
+pub const PROVIDER: &str = "eswin-eic7700-storage";
+
+const MOUNT_TIMEOUT: Duration = Duration::from_secs(10);
+const MOUNT_INTERVAL: Duration = Duration::from_secs(1);
+
+const TARGET_BOOTLOADER: (&str, &str) = ("bootloader", "bootloader_ddr5_secboot.bin");
+
+#[derive(Deserialize, Debug, Default)]
+struct EswinEic7700StorageParameters {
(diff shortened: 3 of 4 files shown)


sjoerdsimons:
> > @sjoerdsimons FYI
> 
> A few items to discuss/pending:
> 
> * Named as `hifive-p550-storage`. As you already mentioned, this is SoC ESWIN EIC7700X specific, so I'll rename it as `eswin-eic7700-storage`. Agreed?

yes please ; I wonder if it willapply to more eswin SoCs but i guess this is a reasonable start. (we can always rename later)

> * Boardswarm expects the storage mountpoint to be under the Systemd StateDirectory (or the user's state directory otherwise). Should we have an entry in the configuration `server.conf` file (server sertion)?

Need to read you code first for that i think; As a first gut feeling having it in the provider config would make more  then in the server... Unless this is something more generic

> * The `commit` implementation is still pending for the volume. What would be the expected behavior there?
>   * Wait for the filesystem to be unmounted? Sync the filesystem?

as noted tersely in the grpc protocol:

 // Commit all data in an uploader specific manner; e.g. may trigger a USB reset for DFU devices

Basically it's meant to tell the volume "I'm done, go do your thing". So i guess for this provider it depends on what triggers the reboot/reset. If that's un unmount/eject do that. But in any case flushing to the filesystem seems a bare minimum.

boardswarm/Cargo.toml
```diff
@@ -45,3 +45,6 @@ reqwest = { version = "0.12.15", default-features = false, features = ["rustls-t
 strum = { version = "0.27.1", features = ["derive"] }
 strum_macros = "0.27.1"
 regex = "1.11.1"
+systemd-directories = "0.1.1"
+proc-mounts = "0.3.0"
+dirs = "6.0.0"
```
sjoerdsimons, line 50: 2 crates just to get env variables?

boardswarm/src/hifive_p550_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+        .get("udev.ID_FS_UUID")
+        .ok_or_else(|| anyhow::anyhow!("Failed to read udev property ID_FS_UUID"))?;
+
+    let state_dir = fs::canonicalize(
+        systemd_directories::state_dir()
+            .or(dirs::state_dir())
+            .ok_or_else(|| anyhow::anyhow!("Failed to get systemd state directory"))?,
+    )
+    .await?;
+    let path = format!("{}/{}", state_dir.to_string_lossy(), uuid);
```
sjoerdsimons, line 81: don't make paths like this; use PathBuf::join

boardswarm/src/hifive_p550_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+
+#[derive(Debug)]
+pub struct HifiveP550StorageVolume {
+    path: String,
+    targets: Vec<VolumeTargetInfo>,
+}
+
+impl HifiveP550StorageVolume {
+    fn new(path: String) -> Self {
+        let targets = TARGETS
```
sjoerdsimons, line 118: that's a lot of generalisation for soemthing with one volume name and target

boardswarm/src/hifive_p550_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+                seekable: false,
+                size: None,
+                blocksize: None,
+            })
+            .collect();
+        Self { path, targets }
+    }
+
+    fn target_path(&self, target: &str) -> Result<String, VolumeError> {
+        let (_, path) = TARGETS
```
sjoerdsimons, line 133: you have one target name and one path; just simply this :)

boardswarm/src/hifive_p550_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    async fn open(
+        &self,
+        target: &str,
+        _length: Option<u64>,
+    ) -> Result<(VolumeTargetInfo, Box<dyn VolumeTarget>), VolumeError> {
+        if let Some(volume_target) = self.targets.iter().find(|t| t.name == target) {
+            let path = self.target_path(target)?;
+            let file = fs::File::create(path).await.unwrap();
```
sjoerdsimons, line 160: return an error; don't unwrap

boardswarm/src/hifive_p550_storage.rs
```diff
@@ -0,0 +1,209 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // TODO make more meaningful
+        f.debug_struct("HifiveP550StorageTarget")
+            .finish_non_exhaustive()
+    }
+}
+
+#[async_trait::async_trait]
+impl VolumeTarget for HifiveP550StorageTarget {
+    async fn write(&mut self, data: Bytes, _offset: u64, completion: crate::WriteCompletion) {
```
sjoerdsimons, line 191: you can't ignore the offset; either verify the offset or seek to it (and set to `seekable: true`)

sjoerdsimons (changes requested):

boardswarm/src/hifive_p550_storage.rs
adalessandro: Done.

boardswarm/src/hifive_p550_storage.rs
adalessandro: Done.

boardswarm/Cargo.toml
adalessandro: Yeah, another sign this might be completely wrong and should be a server configuration instead.

boardswarm/src/hifive_p550_storage.rs
adalessandro: Done.

adalessandro:
> > > The user running boardswarm need to have access to the directory where
> > > the storage device is mounted. Boardswarm expects this mountpoint to be
> > > located at the Systemd StateDirectoy (e.g. /var/lib/boardswarm/)
> > > identified by its UUID.
> > > For example, a udev rule like the following could be added to automount
> > > the device:
> > > ```
> > > $ cat /etc/udev/rules.d/99-boardswarm.rules
> > > ACTION=="add", SUBSYSTEMS=="usb", SUBSYSTEM=="block", \
> > >     ENV{ID_VENDOR}=="ESWIN", ENV{ID_MODEL}=="WIN-2030", TAG+="systemd", \
> > >     RUN{program}+="/usr/bin/systemd-mount --owner=boardswarm --no-block --automount=no --collect $devnode /var/lib/boardswarm/$env{ID_FS_UUID}"
> > > ```
> > 
> > 
> > Is that UUID unique for these devices or would it potentially clash?
> 
> Good question, it would depend on how the fs was created or even the uuid set to it, so there's no guarantee that would be unique.

FWIW, I've replaced it by `ID_PATH` instead of `ID_FS_UUID`.

boardswarm/Cargo.toml
sjoerdsimons: not necessarilly ; but might be simpler to just implement it directly. though maybe erroring on simplicity and just require a absolute path in the provider config may be best. THe user will need to configure the mounting to happen there anyway

adalessandro: Added as a provider parameter `mountpoint: String`
```
struct EswinEic7700StorageParameters {
    #[serde(rename = "match")]
    match_: HashMap<String, String>,
    mountpoint: String,
}
```

adalessandro: Example configuration:
```
  - name: eswin-eic7700-storage
    provider: eswin-eic7700-storage
    parameters:                                                                 
      match:                                                                    
        udev.ID_VENDOR: "ESWIN"                                                 
        udev.ID_MODEL: "WIN-2030"                                               
      mountpoint: "/var/run/boardswarm/eswin-eic7700-storage"
```

sjoerdsimons (approved):

PR #214, opened 2025-08-07
Add python scripting page

mwelchuk:
This may not be the most elegant way to perform testing using Boardswarm, however it did allow me to fairly efficiently run some one-off testing that I needed to perform and I can see myself taking this approach again when I have some tasks that I wish to perform on boards connected to boardswarm, unless of course a better alternative is shown to be in the meantime.


mwelchuk:
Split out from #180.

PR #233, opened 2025-09-25
    boardswarm: mediatek_brom: Add bootstrap address parameter

adalessandro:
Depends on https://github.com/boardswarm/mediatek-brom/pull/4

```
    boardswarm: mediatek_brom: Add bootstrap address parameter
    
    Mediatek platforms might use different bootstrap addresses to load/execute
    the Download Agent (DA). The mediatek-brom crate now supports getting the
    right address from the retrieved hardware code information.
    
    Let's use this, but also add an address parameter in the mediatek-brom
    provider to make it configurable.
```


sjoerdsimons:
Fwiw when doing a PR like this; you can temporarily depend on a git version of the dependent crate so it's actually buildable/CI'able :) Should be in draft mode though normally, though i've just published mediatek-brom 0.2.0 so at this point you can update

PR #276, opened 2026-04-04
udev: skip device Devantech USB Devices

obbardc:
Since the Devantech USB devices are not consumed by boardswarm, instead it is usually used by a pdudaemon instance, add an exception to udev rules to ignore the devices.

See the sister PR for pdudaemon at: https://github.com/pdudaemon/pdudaemon/pull/218/

### DIFF
--- a/boardswarm/share/99-boardswarm.rules
+++ b/boardswarm/share/99-boardswarm.rules
@@ -1,6 +1,14 @@
+# Skip Devantech USB devices as these are usually consumed by pdudaemon.
+# The Devantech USB product range uses the Microchip VID & a sub-licenced PID
+# specific to all Devantech products.
+# Warning: This catch-all rule will match more than the USB relay range.
+ACTION=="add", SUBSYSTEM=="tty", ENV{ID_BUS}=="usb", ATTRS{idVendor}=="04d8", ATTRS{idProduct}=="ffee", GOTO="boardswarm_end"
+
 # USB Serial devices
 ACTION=="add", SUBSYSTEM=="tty", ENV{ID_BUS}=="usb", GROUP="boardswarm"
 
 # TI DFU devices
 ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6162", GROUP="boardswarm"
 ACTION=="add", SUBSYSTEM=="usb", ATTRS{idVendor}=="0451", ATTRS{idProduct}=="6165", GROUP="boardswarm"
+
+LABEL="boardswarm_end"


boardswarm/share/99-boardswarm.rules
```diff
@@ -1,6 +1,12 @@
+# Ignore Devantech USB Relay devices
+# Skip Devantech USB-RLY08C
+ACTION=="add", SUBSYSTEM=="tty", ENV{ID_BUS}=="usb", ENV{ID_VENDOR_ID}=="04d8", ENV{ID_MODEL_ID}=="ffee", GOTO="boardswarm_end"
```
sjoerdsimons: That VID is a MicroChip one. Is the PID actually Devantech specific? If so is it for all their models or just one ? (i should have a RLY02 lying around, i can check that tomorrow). 

Note that I'm not entirely against this.. Though boardswarm shouldn't touch it anyway (iirc it's a ttyACMx device, and boardswarm by design only opens up serial devices when it's asked to or knows for sure it has a provider for it).  Did you see any issues with it ?. As a guess the issue you hit is that you want it to be owned by pdudaemon and not have the rule below for usb apply? 

Wonder if we can do this better in general to make it easy to opt out devices here rather then just adding them all in this rules file :)

obbardc: I am not sure if the PID is devantech specific, but it appears like this when connecting which makes me lean to it being a generic PID from the microchip stack:

```
usb 1-1.1.4.4.3: new full-speed USB device number 26 using xhci_hcd
usb 1-1.1.4.4.3: New USB device found, idVendor=04d8, idProduct=ffee, bcdDevice= 1.00
usb 1-1.1.4.4.3: New USB device strings: Mfr=1, Product=2, SerialNumber=3
usb 1-1.1.4.4.3: Product: USB-RLY08.
usb 1-1.1.4.4.3: Manufacturer: Devantech Ltd.
usb 1-1.1.4.4.3: SerialNumber: 16913xxx
cdc_acm 1-1.1.4.4.3:1.0: ttyACM0: USB ACM device
```

Perhaps we could match on `Manufacturer` instead, if we go this route.

The sample udev rules here changed the owner to boardswarm and the pdudaemon rules then didn't work.

It'd be great if we could filter these unsupported devices out :-). I am happy to rework this to be more generic, but not sure exactly how.

sjoerdsimons: Right so the RLY02 is:
```
usb 1-1: new full-speed USB device number 4 using xhci_hcd
usb 1-1: New USB device found, idVendor=04d8, idProduct=ffee, bcdDevice= 1.00
usb 1-1: New USB device strings: Mfr=1, Product=2, SerialNumber=3
usb 1-1: Product: USB-RLY02.
usb 1-1: Manufacturer: Devantech Ltd.
usb 1-1: SerialNumber: 00015636
```

It seems microchip does offer people to use PIDs with their VID; so hopefully this PID/VID is unique for devantech. 

Practically i wouldn't be against filtering this out; But it doesn't really solve the general problem as there are lot of serial devices that probably want to be used for !boardswarm; But there is no way we can list these all.

Note that you say "unsupported devices". It's not actually unsupported you can use the RLY serial port over boardswarm just fine and send the bytes required to make it switch.. It's just practically not that useful :)

obbardc: > It seems microchip does offer people to use PIDs with their VID; so hopefully this PID/VID is unique for devantech.

I will email devantech and ask, I will include you in CC.

> Practically i wouldn't be against filtering this out; But it doesn't really solve the general problem as there are lot of serial devices that probably want to be used for !boardswarm; But there is no way we can list these all.

Yeah, I am not sure how we would do that really. Open to suggestions, or maybe we merge this and open an issue for future us to figure out?

> Note that you say "unsupported devices". It's not actually unsupported you can use the RLY serial port over boardswarm just fine and send the bytes required to make it switch.. It's just practically not that useful :)

Very pedantic, you know exactly what I mean ;-).

obbardc: Devantech support confirmed on 22 April 2026 that the Product ID is sub-licensed to them as a unique identifier, but that the same PID is reused across multiple Devantech USB modules rather than being assigned per individual product.

I updated the udev rule to match against all Devantech USB devices, but I think that is fine since they are mainly USB relay modules which are compatible with pdudaemon.

I could add rules for all supported products and even ask Devantech for a list of Product attrs for their USB relay range, but I think that's a bit excessive and matching on all devices is fine for now.

Unless you have any other suggestions to the general problem we could implement soon, I think it'd be great if you could consider merging this now and opening an issue to figure out solving the "general" problem.

sjoerdsimons: Yeah i think this is fine

PR #277, opened 2026-04-05
Add uhubctl actuator provider

obbardc:
Add a new provider to integrate uhubctl and control USB hub port power.
Each configured port is registered as an actuator supporting "on" and
"off" modes, which invokes `uhubctl -l <location> -p <port> -a <on|off>`.

The provider is configured with a hub location, an optional path to the
uhubctl binary and a list of ports with descriptive names.

### DIFF
--- a/boardswarm/Dockerfile
+++ b/boardswarm/Dockerfile
@@ -33,6 +33,7 @@ LABEL org.opencontainers.image.source="https://github.com/boardswarm/boardswarm"
 RUN apt update && \
     apt install --yes \
       ca-certificates \
+      uhubctl \
       libudev1 \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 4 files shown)


obbardc:
I doubt that integrating this tool is the right approach, I am assuming @sjoerdsimons would rather a rust implementation :p

sjoerdsimons:
> I doubt that integrating this tool is the right approach, I am assuming @sjoerdsimons would rather a rust implementation :p

Yup calling out to things seems more of a pdudaemon thing ;) Especially as i'ts just on/off

obbardc:
> > I doubt that integrating this tool is the right approach, I am assuming @sjoerdsimons would rather a rust implementation :p
> 
> Yup calling out to things seems more of a pdudaemon thing ;) Especially as i'ts just on/off

Huh, I never thought about that. I guess it would make sense to integrate it into pdudaemon as a driver instead!

I will leave this PR open for now, until I get time to implement it in pdudaemon, since I am currently using this for a setup.

PR #283, opened 2026-04-15
cli: add interactive device picker in UI mode

obbardc:
When `boardswarm-cli ui` is invoked without a device argument, display a TUI list of all devices available on the server. The user can navigate with arrow keys or j/k, confirm with Enter and abort with q/Escape.

### DIFF
--- a/boardswarm-cli/src/ui.rs
+++ b/boardswarm-cli/src/ui.rs
@@ -1,13 +1,17 @@
 use std::task::Poll;
 use std::{num::ParseIntError, str::FromStr};
 
+use boardswarm_client::client::Boardswarm;
+use boardswarm_protocol::ItemType;
 use bytes::Bytes;
 use futures::{Stream, StreamExt, pin_mut, ready};
 use ratatui::{
     Terminal as TuiTerminal,
     backend::CrosstermBackend,
+    crossterm::event::{Event, KeyCode, KeyEventKind},
     layout::{Rect, Size},
-    widgets::{Block, Borders},
+    style::Style,
+    widgets::{Block, Borders, List, ListItem, ListState},
 };
 use thiserror::Error;
 use tokio::io::{AsyncRead, AsyncReadExt};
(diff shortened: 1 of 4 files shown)


boardswarm-cli/src/ui.rs
```diff
@@ -238,6 +242,78 @@ where
...
+}
+
+fn run_device_picker(
+    terminal: &mut ratatui::DefaultTerminal,
+    items: &[boardswarm_protocol::Item],
+) -> anyhow::Result<u64> {
+    let mut state = ListState::default();
+    state.select(Some(0));
+
+    loop {
```
obbardc: I am not sure a loop is right here.

boardswarm-cli/src/ui.rs
```diff
@@ -238,6 +242,78 @@ where
...
+        terminal.draw(|f| {
+            let area = f.area();
+            let list_items: Vec<ListItem> = items
```
obbardc: sounds silly, but we should handle the case when the device list is bigger than the terminal

obbardc: I tested it, works fine !

sjoerdsimons: so this is not an issue then? (if so close your self-comment :p )

boardswarm-cli/src/ui.rs
```diff
@@ -238,6 +242,78 @@ where
...
+pub async fn select_device(
+    mut boardswarm: Boardswarm,
+) -> anyhow::Result<boardswarm_client::device::Device> {
+    let items = boardswarm.list(ItemType::Device).await?;
+
+    if items.is_empty() {
+        anyhow::bail!("No devices available on the server");
+    }
+
+    let mut terminal = ratatui::init();
```
sjoerdsimons: It seems rather weird to have this init ratatui for selection and then the main ui re-inits ratatui again; the device selection should really be a state of the UI. Which would also open up the possibility of changing devices in future

obbardc: Sure, I will improve this code. For now I just threw it up to get your thoughts on if it's a good idea to improve the UI :-).

I think adding a hotkey to change device is a nice idea. I will see if I can do that now or create an issue if not.

obbardc: I did this change. I am sure it's a bit spaghetti and you may have better idea of how I can implement the state machine?

I left the original commit in, but I think it would make sense to split into three commits:
- add UI state machine
- add device picker UI to state machine
- add ability to switch device at runtime

obbardc: @sjoerdsimons any general comments on this before I cleanup the commits ? I just rebased on main.

PR #300, opened 2026-07-11
Add --version argument to boardswarm and boardswarm-cli

obbardc:
Add an argument to the boardswarm CLI binaries to report the version.

The version is determined at build time in this order:
1. BOARDSWARM_VERSION environment variable
2. `git describe`; which produces "<crate version> (<commit>[-dirty])"
3. the crate version from Cargo.toml

Also inject the version when building in GitHub actions and when building the boardswarm containers.

### DIFF
--- a/boardswarm-cli/Dockerfile
+++ b/boardswarm-cli/Dockerfile
@@ -1,5 +1,6 @@
 ## stage 1: builder
 FROM rust:slim-trixie AS builder
+ARG BOARDSWARM_VERSION
 ARG BINARY
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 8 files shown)


boardswarm-cli/Dockerfile
```diff
@@ -11,6 +11,11 @@ RUN apt-get update && \
 
 WORKDIR /usr/src/boardswarm
 
+# Version string embedded into the binary (see build.rs). The build context
+# excludes .git, so this is injected by CI instead of derived from git.
+ARG BOARDSWARM_VERSION
+ENV BOARDSWARM_VERSION=${BOARDSWARM_VERSION}
```
obbardc: What happens if this isn't set ?

boardswarm/Dockerfile
```diff
@@ -12,6 +12,11 @@ RUN apt-get update && \
 
 WORKDIR /usr/src/boardswarm
 
+# Version string embedded into the binary (see build.rs). The build context
```
obbardc: What happens if this isn't set ?

